PR #1, opened 2021-06-05
resync feature set with the NFD changes

fromanirh:
This PR brings up to speed with the features we developed for NFD.

Originally RTE was meant as prototype/PoC for the code that would later evolved into a NFD PR.
The plan hasn't changed a bit: we envision NFD as a proper long term support for the topology-aware topology zones reporting.
However, is good to have a up-to-date separate prototype we can use to crosscheck the implementation and to try out new ideas, and to experiment in general.

Hence, we resync back RTE with NFD.

We import back (clone, not vendor) the packages we created for NFD, reusing them unchanged, and we fix the other bits and pieces to integrate with them


Makefile
```diff
@@ -23,48 +23,38 @@ govet:
...
 push: image
 	@echo "pushing image"
 	$(RUNTIME) push quay.io/$(REPOOWNER)/$(IMAGENAME):$(IMAGETAG)
 
-.PHONY: deploy
-deploy: push
-	@echo "deploying Resource Topology Exporter"
-	kubectl create -f manifests/resource-topology-exporter-ds.yaml
-
-.PHONY: deploy-pod
```
swatisehgal, line 51: Do we not want to have a make target here for deployment of the sample pods requesting sriov devices that are in the manifest directory?

cmd/resource-topology-exporter/main.go
```diff
@@ -0,0 +1,137 @@
...
+	if err != nil {
+		log.Fatalf("failed to execute: %v", err)
+	}
+}
+
+const helpTemplate string = `{{.ProgramName}}
+
+  Usage:
+  {{.ProgramName}} [--no-publish]
+                 [--oneshot | --sleep-interval=<seconds>]
```
swatisehgal, line 39: NIT: fix the indentation here

Makefile
```diff
@@ -1,6 +1,8 @@
 COMMONENVVAR=GOOS=linux GOARCH=amd64
 BUILDENVVAR=CGO_ENABLED=0
+TOPOLOGYAPI_MANIFESTS=https://raw.githubusercontent.com/k8stopologyawareschedwg/noderesourcetopology-api/master/manifests
 
+KUBECTL ?= kubectl
```
swatisehgal, line 5: Maybe better to call it `CLI` or `KUBECLI` instead of `KUBECTL` here?

Makefile
```diff
@@ -58,3 +60,13 @@ test-unit:
 .PHONY: test-e2e
 test-e2e: binaries
 	ginkgo test/e2e
+
+.PHONY: deploy
+deploy:
+	$(KUBECTL) create -f $(TOPOLOGYAPI_MANIFESTS)/crd.yaml
```
swatisehgal, line 66: Good idea to refer to the CRD manifest in the noderesourcetopology-api!

manifests/resource-topology-exporter-ds.yaml
```diff
@@ -54,32 +54,25 @@ spec:
       - name: resource-topology-exporter-container
         image: quay.io/swsehgal/resource-topology-exporter:latest
         command:
-        - /usr/local/bin/resource-topology-exporter
-        - --watch-namespace=rte
+        - /bin/resource-topology-exporter
+        - --export-namespace=default
```
swatisehgal, line 48: I see in the  manifest directory we have test pods requesting SR-IOV devices in the `rte` namespace, so think it would be better to provide `--watch-namespace=rte` here as well.

manifests/resource-topology-exporter-ds.yaml
```diff
@@ -48,11 +39,10 @@ spec:
       labels:
         name: resource-topology
     spec:
-      hostNetwork: true
-      serviceAccountName: rte-account
+      serviceAccountName: rte
       containers:
       - name: resource-topology-exporter-container
-        image: quay.io/swsehgal/resource-topology-exporter:latest
+        image: ${RTE_CONTAINER_IMAGE}
```
swatisehgal, line 45: Maybe like NFD we could also have a template files where ${RTE_CONTAINER_IMAGE} replaced by appropriate name everytime `make image` is executed? Of course not needed here but is a nice to have, could be done in a separate PR.

manifests/resource-topology-exporter-ds.yaml
```diff
@@ -29,7 +29,7 @@ roleRef:
 apiVersion: apps/v1
 kind: DaemonSet
 metadata:
-  name: resource-topology-exporter-ds
+  name: rte
```
swatisehgal, line 32: I think it is better to leave it as `resource-topology-exporter-ds` here because we have been using `rte` as namespace in a lot of places. Even though I know it wouldn't cause any confusion but I don't want it to appear to the user might thing that the daemon set name has something to do with the watch-namespace.

swatisehgal (commented):
I have tested this on my local environment and can confirm that this works as expected. A few minor comments and observations but other than that it looks good to me!

cmd/resource-topology-exporter/main.go
fromanirh: done

Makefile
fromanirh: used `KUBECLI`

Makefile
fromanirh: :)

manifests/resource-topology-exporter-ds.yaml
fromanirh: sure, added

manifests/resource-topology-exporter-ds.yaml
fromanirh: yes, I prefer to postpone because is not yet clear where the prebuilt image is going to sit, and I think we should avoid personal repos if possible.

manifests/resource-topology-exporter-ds.yaml
fromanirh: sure, fixed

swatisehgal:
/lgtm
/approve

Thanks for making all the updates and for your contribution in completing the first step of bringing Resource Topology Exporter upto speed with NFD!

PR #2, opened 2021-06-12
cmd: allow to override sysfs mount point

fromanirh:
allow to use `/host-sys` and not force `/host/sys` (we can still do the latter).
Import ghw package which allows so.


swatisehgal:
/lgtm

Thanks!

PR #3, opened 2021-06-13
Add re-sync git action

Tal-or:
This workflow rebase the u/s repo into its d/s mirror in order to keep them in sync.
In addition, in case of conflicts, the job will exit with an error and requires manual intervention.
Job's interval: After every new push.

### DIFF
--- a/.github/workflows/resync-master.yaml
+++ b/.github/workflows/resync-master.yaml
@@ -1,0 +1,44 @@
+# This workflow will rebase the u/s repo onto it's d/s mirror in order to keep them in sync
+
+name: resync action
+
+on:
+  push:
+    branches: [ master ]
+
+jobs:
+  resync:
+    name: resync job
+    runs-on: ubuntu-latest
+
+    steps:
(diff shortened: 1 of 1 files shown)


fromanirh:
good idea, but u/s projects don't push themselves to their d/s consumers. It's d/s that pull u/s. So we should not merge this PR.

Tal-or:
> good idea, but u/s projects don't push themselves to their d/s consumers. It's d/s that pull u/s. So we should not merge this PR.

I can implement the same on the d/s side.
But what will be the trigger for the action to run?

fromanirh:
> > good idea, but u/s projects don't push themselves to their d/s consumers. It's d/s that pull u/s. So we should not merge this PR.
> 
> I can implement the same on the d/s side.
> But what will be the trigger for the action to run?

A periodic trigger (like daily, but probably even less frequent) should be fine

Tal-or:
Close this PR. Due to some unresolved issues, we will maintain manual sync for now.

PR #5, opened 2021-06-21
Add smart polling for RTE

fromanirh:
The "smart polling" idea is all about have more polling triggers than just timer.
We know -and we can pretty much depend on- that kubelet rewrites it checkpoints for resource manager each time a resource
allocation is performed. When this happens, something changed and we can poll again the podresources API to learn what.

This allows much faster response time without the need of having a very frequent poll interval, which is too taxing on steady state.

To quickly learn about the checkpoint changes, we use the standard `fsnotify` linux API using the go package.

The only drawback of this (optional) approach is that RTE now requires RO access to the while kubelet state directory(ies).

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -36,7 +36,7 @@ jobs:
 
     - name: generate manifests
       run: |
-        RTE_CONTAINER_IMAGE=${RTE_CONTAINER_IMAGE} make gen-manifests > rte-e2e.yaml
+        RTE_CONTAINER_IMAGE=${RTE_CONTAINER_IMAGE} RTE_POLL_INTERVAL=10s make gen-manifests | tee rte-e2e.yaml
 
     - name: create K8S kind cluster
       run: |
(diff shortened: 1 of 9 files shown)


.github/workflows/e2e.yml
```diff
@@ -48,6 +48,13 @@ jobs:
       run: |
         kubectl create -f rte-e2e.yaml
 
+    - name: cluster info
+      run: |
+        kubectl get nodes
+        kubectl get pods -A
+        kubectl describe pod -l name=resource-topology || :
```
Tal-or, line 55: Why avoid the  ```kubectl describe pod -l name=resource-topology``` call from return with error in case of failure? can the pod has a different name?

fromanirh: because we have a race on container start, and I didn't want to complicate the steps adding wais for the container to be Ready - we would need to add and handle timeouts in this case.

cmd/resource-topology-exporter/main.go
```diff
@@ -41,6 +41,7 @@ const helpTemplate string = `{{.ProgramName}}
 			[--export-namespace=<namespace>]
 			[--watch-namespace=<namespace>]
 			[--sysfs=<mountpoint>]
+			[--kubelet-state-dir=<path> --kubelet-state-dir=<path>]
```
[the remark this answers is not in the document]
fromanirh: It's a repeating option (http://docopt.org/). Let me check if looks clearer using ellipsis.

PR #6, opened 2021-06-22
Add and use a minion container to learn about the shared cpu pool

fromanirh:
Until the kubernetes issue https://github.com/kubernetes/kubernetes/issues/102190
is fixed, we can't learn easily from the podresources API which
cpus belong to the shared pool and which cpus are assigned to exclusive
pools. This makes the resource reporting inconsistent.

A possible workaround is to explicitely design a container whose
cpu resources belong to the shared pool. Knowing that, RTE can safely
compute which cpus pertain to which pool when scanning the resources
allocated to each container.

So, we add a minion pause container, which we know for sure
1. it always runs
2. where it runs (namespace, pod name, container name)
3. gets cpus from the shared pool

Then, we point the RTE to it, so it is enabled to fix the cpu reporting.

### DIFF
--- a/cmd/resource-topology-exporter/main.go
+++ b/cmd/resource-topology-exporter/main.go
@@ -21,12 +21,12 @@ const (
 )
 
 func main() {
-	nrtupdaterArgs, resourcemonitorArgs, err := argsParse(os.Args[1:])
+	nrtupdaterArgs, resourcemonitorArgs, rteArgs, err := argsParse(os.Args[1:])
 	if err != nil {
 		log.Fatalf("failed to parse command line: %v", err)
 	}
 
-	err = resourcetopologyexporter.Execute(nrtupdaterArgs, resourcemonitorArgs)
+	err = resourcetopologyexporter.Execute(nrtupdaterArgs, resourcemonitorArgs, rteArgs)
 	if err != nil {
 		log.Fatalf("failed to execute: %v", err)
 	}
@@ -43,6 +43,7 @@ const helpTemplate string = `{{.ProgramName}}
 			[--sysfs=<mountpoint>]
 			[--kubelet-state-dir=<path>...]
 			[--kubelet-config-file=<path>]
+			[--reference-container=<spec>]
 
   {{.ProgramName}} -h | --help
   {{.ProgramName}} --version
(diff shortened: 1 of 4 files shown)


Tal-or:
besides the conflicts /lgtm

cmd/resource-topology-exporter/main.go
```diff
@@ -63,7 +64,8 @@ const helpTemplate string = `{{.ProgramName}}
                                   [Default: /kubeletstate/config.yaml]
   --kubelet-state-dir=<path>...   Kubelet state directory (RO access needed), for smart polling.
   --podresources-socket=<path>    Pod Resource Socket path to use.
-                                  [Default: /podresources/kubelet.sock]`
+                                  [Default: /podresources/kubelet.sock]
+  --reference-container=<spec>    Reference container (namespace/podname/containername)`
```
swatisehgal, line 68: We should add a better description for the user here. Since there are no default values here it might appear that this is an optional flag which is not the case as we always need to determine the shared pool. So lets mention explicitly that if values are not provided here, environment variables will be used and we should have some default values in hack/get-manifest-ds.sh.

manifests/resource-topology-exporter-ds.yaml
```diff
@@ -56,6 +56,16 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: spec.nodeName
+        - name: REFERENCE_NAMESPACE
```
swatisehgal, line 59: Don't we need ${REFERENCE_NAMESPACE}, ${REFERENCE_POD_NAME} and ${REFERENCE_CONTAINER_NAME} here?

swatisehgal (commented):
Thanks for the PR. Did a first pass, really like the way this has been implemented with the filteringClient abstraction! A few minor comments but I think we need to add more tests to be able to highlight the exact problem where guaranteed pods with non integral requests need to be deployed and handled appropriately.

cmd/resource-topology-exporter/main.go
fromanirh: good point, will add.

manifests/resource-topology-exporter-ds.yaml
fromanirh: no, because this is the name of the env vars being created

fromanirh:
regarding e2e tests, I'm planning to add a bunch of them in a followup PR

swatisehgal:
> regarding e2e tests, I'm planning to add a bunch of them in a followup PR

Sure, perfectly okay to add more tests in a follow-up PR.

manifests/resource-topology-exporter-ds.yaml
swatisehgal: Ah, okay. Sorry missed that these are the environment variable KEY values.

PR #8, opened 2021-07-05
Backport excludelist

Tal-or:
Backport the exclude-list feature from NFD.
The original PR for NFD: kubernetes-sigs/node-feature-discovery#545

For the E2E there is some problem with creating the ConfigMap during the e2e test.
The ConfigMap should be created on the cluster before the RTE container is running (the RTE container consumes the ConfigMap). this is not the case because the RTE container is up and running before the test even starts (it was created by ```kubectl create -f``` in previous steps of the CI)

Due to that, the E2E test for the ConfigMap will be added on a separate PR

### DIFF
--- a/cmd/resource-topology-exporter/main.go
+++ b/cmd/resource-topology-exporter/main.go
@@ -3,12 +3,14 @@ package main
 import (
 	"bytes"
 	"fmt"
+	"io/ioutil"
 	"log"
 	"os"
 	"text/template"
 	"time"
 
 	"github.com/docopt/docopt-go"
+	"sigs.k8s.io/yaml"
 
 	"github.com/k8stopologyawareschedwg/resource-topology-exporter/pkg/nrtupdater"
 	"github.com/k8stopologyawareschedwg/resource-topology-exporter/pkg/resourcemonitor"
@@ -45,6 +47,7 @@ const helpTemplate string = `{{.ProgramName}}
 			[--kubelet-state-dir=<path>...]
 			[--kubelet-config-file=<path>]
 			[--reference-container=<spec>]
+			[--exclude-list-config=<path>]
 
   {{.ProgramName}} -h | --help
   {{.ProgramName}} --version
@@ -70,8 +73,10 @@ const helpTemplate string = `{{.ProgramName}}
   --reference-container=<spec>    Reference container, used to learn about the shared cpu pool
                                   See: https://github.com/kubernetes/kubernetes/issues/102190
                                   format of spec is namespace/podname/containername.
-				  Alternatively, you can use the env vars
-				  REFERENCE_NAMESPACE, REFERENCE_POD_NAME, REFERENCE_CONTAINER_NAME.`
+                                  Alternatively, you can use the env vars
+				                  REFERENCE_NAMESPACE, REFERENCE_POD_NAME, REFERENCE_CONTAINER_NAME.
+  --exclude-list-config=<path>    Exclude resources list file path.
+                                  [Default:/etc/resource-topology-exporter-config/exclude-list-config.yaml]`
 
 func getUsage() (string, error) {
 	var helpBuffer bytes.Buffer
@@ -157,5 +162,31 @@ func argsParse(argv []string) (nrtupdater.Args, resourcemonitor.Args, resourceto
 		rteArgs.ReferenceContainer = resourcetopologyexporter.ContainerIdentFromEnv()
 	}
 
+	if excludeListConfigMapPath, ok := arguments["--exclude-list-config"].(string); ok {
+		resourcemonitorArgs.ExcludeList, err = getExcludeListFromConfigMap(excludeListConfigMapPath)
+		if err != nil {
+			log.Fatalf("error getting exclude list from the configutarion: %v", err)
+		}
+	}
 	return nrtupdaterArgs, resourcemonitorArgs, rteArgs, nil
+}
+
+func getExcludeListFromConfigMap(configMapPath string) (resourcemonitor.ResourceExcludeList, error) {
+	excludeList := resourcemonitor.ResourceExcludeList{}
+
+	config, err := ioutil.ReadFile(configMapPath)
+	if err != nil {
+		// ConfigMap is optional
+		if os.IsNotExist(err) {
+			log.Printf("Info: couldn't find configuration under %v", configMapPath)
+			return excludeList, nil
+		}
+		return excludeList, err
+	}
+
+	err = yaml.Unmarshal(config, &excludeList)
+	if err != nil {
+		return excludeList, err
+	}
+	return excludeList, nil
 }
--- a/manifests/resource-topology-exporter-config.yaml
+++ b/manifests/resource-topology-exporter-config.yaml
@@ -1,0 +1,19 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: rte
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: resource-topology-exporter-config
+  namespace: rte
+data:
+  exclude-list-config.yaml: |
+    # key = node name, value = list of resources to be excluded.
+    # use * to exclude from all nodes.
+    # an example for how the exclude list should looks like
+    # excludelist:
+    #   node1: [cpu]
+    #   node2: [memory, example/deviceA]
+    #   *: [cpu]
--- a/pkg/resourcemonitor/noderesourcesaggregator_test.go
+++ b/pkg/resourcemonitor/noderesourcesaggregator_test.go
@@ -26,6 +26,7 @@ import (
 
 	cmp "github.com/google/go-cmp/cmp"
 	. "github.com/smartystreets/goconvey/convey"
+	"github.com/stretchr/testify/assert"
 	"k8s.io/apimachinery/pkg/util/intstr"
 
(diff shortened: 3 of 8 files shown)


pkg/resourcemonitor/noderesourcesaggregator_test.go
```diff
@@ -227,6 +227,17 @@ func TestResourcesAggregator(t *testing.T) {
 						},
 					},
 				},
+				&v1.ContainerDevices{
+					ResourceName: "fake.io/resourceToBeExcluded",
```
swatisehgal, line 231: The resourceName "fake.io/resourceToBeExcluded" is not in the config file?

swatisehgal (commented):
Took a pass at it, will take a deeper look once again later. The E2E tests are currently failing as can be seen in the CI, could you fix that please.

swatisehgal:
Another NIT, in the commit message please update `NDF` -> `NFD`

manifests/resource-topology-exporter-config.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: rte
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: resource-topology-exporter-config
```
fromanirh, line 9: note that the `-config` suffix is redundant for stuff we put under `/etc`, because `/etc` is exactly the filesystem location which holds configuration

pkg/resourcemonitor/noderesourcesaggregator_test.go
fromanirh: If this is intended, please add a comment to document so.

test/e2e/rte.go
```diff
@@ -81,4 +89,32 @@ var _ = ginkgo.Describe("[RTE] Node topology updater", func() {
 			}, time.Minute, 5*time.Second).Should(gomega.BeTrue())
 		})
 	})
+
+	ginkgo.AfterEach(func() {
```
fromanirh, line 93: so this will run after each test, and now it's fine, but is going to be fine in the future? I'd just move this cleanup inside the tests which create configmaps.

test/e2e/rte.go
```diff
@@ -81,4 +89,32 @@ var _ = ginkgo.Describe("[RTE] Node topology updater", func() {
...
+		err := f.ClientSet.CoreV1().ConfigMaps("default").Delete(context.TODO(), TopologyUpdaterConfigMapName, metav1.DeleteOptions{})
```
fromanirh, line 94: the namespace looks suspicious - we don't create the configmap in the "default" namespace

test/e2e/rte.go
```diff
@@ -81,4 +89,32 @@ var _ = ginkgo.Describe("[RTE] Node topology updater", func() {
...
+		if err != nil {
+			framework.Logf("failed to delete topology updater ConfigMap: %v", err)
```
fromanirh, line 96: this should be a test failure

test/e2e/rte.go
```diff
@@ -81,4 +89,32 @@ var _ = ginkgo.Describe("[RTE] Node topology updater", func() {
...
+	cm := &v1.ConfigMap {
+		ObjectMeta: metav1.ObjectMeta {
+			Name: TopologyUpdaterConfigMapName,
+		},
+		Data: map[string]string{
+			"exclude-list-config.yaml": string(eJSONBytes[:]),
+		},
+	}
+
+	cm, err = cs.CoreV1().ConfigMaps(ns).Create(context.TODO(), cm, metav1.CreateOptions{})
```
fromanirh, line 115: maybe just:
```
_, err = cs.CoreV1().ConfigMaps(ns).Create(context.TODO(), cm, metav1.CreateOptions{})
return err
```
?

cmd/resource-topology-exporter/main.go
```diff
@@ -154,5 +157,31 @@ func argsParse(argv []string) (nrtupdater.Args, resourcemonitor.Args, resourceto
 		rteArgs.ReferenceContainer = resourcetopologyexporter.ContainerIdentFromEnv()
 	}
 
+	resourcemonitorArgs.ExcludeList, err = getExcludeListFromConfigMap(excludeListConfigMapPath)
```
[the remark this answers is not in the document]
Tal-or: Why do we need to allow the user to select the ConfiMap's mount path?
Only its content should change

pkg/resourcemonitor/noderesourcesaggregator_test.go
Tal-or: What config file? if you mean to the on under ```test/data/excludelist.yaml``` I added it by mistake (for a different test) and I shall remove it.

Also this resource is being excluded under ```resAggr.Aggregate()``` [func](https://github.com/k8stopologyawareschedwg/resource-topology-exporter/pull/8/files#diff-ba5f7b85b8aa30feceac36d7be94496487dae04853c1241ce26807a875146045R376)

cmd/resource-topology-exporter/main.go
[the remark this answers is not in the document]
fromanirh: the manifests can (and should) keep the current value. The hardcoded path should be overridable like any other option the tool receives.

manifests/resource-topology-exporter-config.yaml
Tal-or: I basically followed conventions from other projects like topology-aware-scheduler:
https://github.com/k8stopologyawareschedwg/scheduler-plugins/blob/TopologyAwareSchedulerPerNUMA/manifests/noderesourcetopology/deploy.yaml#L50

test/e2e/rte.go
Tal-or: Basically, for every test, I also create the ConfigMap in the ```BeforeEach()``` as an empty one, then it's possible to update its content inside the tests themselves or inside inner ```BeforeEach()```

fromanirh: If we keep the code symmetric, it's fine to keep the code in `BeforeEach` and `AfterEach`. It's also fine to do a garbage-collection-like thingy and _conditionally_ clean up in `AfterEach`. In general I think it's little worth to create a generic configmap and update it on each test - just easier and simpler to just create the new configmap straight in the test.
last but not least, the default kube framework creates a temp test namespace for each run, so cleanup is done automatically when the suite ends. Sometimes it's worth to clean up explicitely to avoid to pollute the global state, though.

Tal-or: BTW  call the ```f.ClientSet``` from the ```AfterEach()``` scope is not valid (end up with nil pointer) I guess that's because the  framework pointer (```f := framework.NewDefaultFramework("rte")```) does not exist in the ```AfterEach()``` context anymore

Tal-or: Actually, after a closer look, I see we can't create the ConfigMap during the test at all.
The ConfigMap should be created on the cluster **before**  the rte container is running (the RTE container consumes the ConfigMap). this is not the case because the RTE container is up and running, before the test even start (it was created by ```kubectl create -f ``` in previous steps of the CI.
Need to find a better way to test it.

Tal-or:
> is not that big of a deal in this case but in general let's try to avoid unnecessary changes, so the PR is kept minimal.

So I'll have to remove the ConfigMap creation from the e2e for now and settle for unit test

fromanirh:
> > is not that big of a deal in this case but in general let's try to avoid unnecessary changes, so the PR is kept minimal.
> 
> So I'll have to remove the ConfigMap creation from the e2e for now and settle for unit test

Sorry I was not very clear. I was referring to quite a few `gofmt` changes applied to unrelated files. Or that was the first impression.

Tal-or:
I can provide the ```gofmt``` as a separate PR, it's not a big deal. I'm more concerned about how to test the exclude-list as part of the e2e tests

cmd/resource-topology-exporter/main.go
```diff
@@ -70,8 +73,10 @@ const helpTemplate string = `{{.ProgramName}}
   --reference-container=<spec>    Reference container, used to learn about the shared cpu pool
                                   See: https://github.com/kubernetes/kubernetes/issues/102190
                                   format of spec is namespace/podname/containername.
-				  Alternatively, you can use the env vars
-				  REFERENCE_NAMESPACE, REFERENCE_POD_NAME, REFERENCE_CONTAINER_NAME.`
+                                  Alternatively, you can use the env vars
+				                  REFERENCE_NAMESPACE, REFERENCE_POD_NAME, REFERENCE_CONTAINER_NAME.
```
fromanirh, line 77: unnecessary whitespace change

cmd/resource-topology-exporter/main.go
```diff
@@ -157,5 +162,31 @@ func argsParse(argv []string) (nrtupdater.Args, resourcemonitor.Args, resourceto
 		rteArgs.ReferenceContainer = resourcetopologyexporter.ContainerIdentFromEnv()
 	}
 
+	if excludeListConfigMapPath, ok := arguments["--exclude-list-config"].(string); ok {
+		resourcemonitorArgs.ExcludeList, err = getExcludeListFromConfigMap(excludeListConfigMapPath)
+		if err != nil {
+			log.Fatalf("error getting exclude list from the configutarion: %v", err)
```
fromanirh, line 168: typo: configuration

fromanirh (approved):
/lgtm

PR #9, opened 2021-07-13
Add more e2e tests

fromanirh:
we increease the coverage of the topology updating tests, and we add the real rte-specific tests.

### DIFF
--- a/pkg/nrtupdater/nrtupdater.go
+++ b/pkg/nrtupdater/nrtupdater.go
@@ -15,6 +15,15 @@ import (
 	"github.com/k8stopologyawareschedwg/resource-topology-exporter/pkg/dumpobject"
 )
 
+const (
+	AnnotationRTEUpdate = "k8stopoawareschedwg/rte-update"
+)
+
+const (
+	RTEUpdatePeriodic = "periodic"
+	RTEUpdateReactive = "reactive"
+)
+
 var (
 	stdoutLogger = log.New(os.Stdout, "", log.LstdFlags)
 	stderrLogger = log.New(os.Stderr, "", log.LstdFlags)
@@ -33,6 +42,11 @@ type NRTUpdater struct {
 	tmPolicy string
 }
 
+type MonitorInfo struct {
+	Timer bool
+	Zones v1alpha1.ZoneList
+}
+
 func NewNRTUpdater(args Args, policy string) (*NRTUpdater, error) {
 	te := &NRTUpdater{
 		args:     args,
@@ -41,8 +55,15 @@ func NewNRTUpdater(args Args, policy string) (*NRTUpdater, error) {
 	return te, nil
 }
 
-func (te *NRTUpdater) Update(zones v1alpha1.ZoneList) error {
-	stdoutLogger.Printf("update: sending zone: '%s'", dumpobject.DumpObject(zones))
+func updateReason(info MonitorInfo) string {
+	if info.Timer {
+		return RTEUpdatePeriodic
+	}
+	return RTEUpdateReactive
+}
+
+func (te *NRTUpdater) Update(info MonitorInfo) error {
+	stdoutLogger.Printf("update: sending zone: '%s'", dumpobject.DumpObject(info.Zones))
 
 	if te.args.NoPublish {
 		return nil
@@ -58,8 +79,11 @@ func (te *NRTUpdater) Update(zones v1alpha1.ZoneList) error {
 		nrtNew := v1alpha1.NodeResourceTopology{
 			ObjectMeta: metav1.ObjectMeta{
 				Name: te.args.Hostname,
+				Annotations: map[string]string{
+					AnnotationRTEUpdate: updateReason(info),
+				},
 			},
-			Zones:            zones,
+			Zones:            info.Zones,
 			TopologyPolicies: []string{te.tmPolicy},
 		}
 
@@ -76,7 +100,11 @@ func (te *NRTUpdater) Update(zones v1alpha1.ZoneList) error {
 	}
 
 	nrtMutated := nrt.DeepCopy()
-	nrtMutated.Zones = zones
+	if nrtMutated.Annotations == nil {
+		nrtMutated.Annotations = make(map[string]string)
+	}
+	nrtMutated.Annotations[AnnotationRTEUpdate] = updateReason(info)
+	nrtMutated.Zones = info.Zones
 
 	nrtUpdated, err := cli.TopologyV1alpha1().NodeResourceTopologies(te.args.Namespace).Update(context.TODO(), nrtMutated, metav1.UpdateOptions{})
 	if err != nil {
@@ -86,14 +114,14 @@ func (te *NRTUpdater) Update(zones v1alpha1.ZoneList) error {
 	return nil
 }
 
-func (te *NRTUpdater) Run(zonesChannel <-chan v1alpha1.ZoneList) chan<- struct{} {
+func (te *NRTUpdater) Run(infoChannel <-chan MonitorInfo) chan<- struct{} {
 	done := make(chan struct{})
 	go func() {
 		for {
 			select {
-			case zonesValue := <-zonesChannel:
+			case info := <-infoChannel:
 				tsBegin := time.Now()
-				if err := te.Update(zonesValue); err != nil {
+				if err := te.Update(info); err != nil {
 					log.Printf("failed to update: %v", err)
 				}
 				tsEnd := time.Now()
--- a/test/e2e/rte.go
+++ b/test/e2e/rte.go
@@ -35,6 +35,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/kubernetes/test/e2e/framework"
 
+	"github.com/k8stopologyawareschedwg/resource-topology-exporter/pkg/nrtupdater"
 	"github.com/k8stopologyawareschedwg/resource-topology-exporter/test/e2e/utils"
 )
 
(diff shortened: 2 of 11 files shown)


pkg/nrtupdater/nrtupdater.go
```diff
@@ -15,6 +15,15 @@ import (
...
+const (
+	AnnotationRTEUpdate = "k8stopoawareschedwg/rte-update"
+)
+
+const (
+	RTEUpdatePeriodic = "periodic"
+	RTEUpdateReactive = "reactive"
+)
+
 var (
```
swatisehgal, line 27: There was feedback on the NFD PR to switch from using log to klog, looks like there is an inclination towards usingklog and we should consider doing that in RTE as well. Could be a completely separate PR.

fromanirh: ack, let's address indeed.

test/e2e/topology_updater.go
```diff
@@ -41,35 +44,154 @@ import (
...
+			finalAllocRes := allocatableResourceListFromNodeResourceTopology(finalNodeTopo)
+			if len(initialAllocRes) == 0 || len(finalAllocRes) == 0 {
+				ginkgo.Fail(fmt.Sprintf("failed to find allocatable resources from node topology initial=%v final=%v", initialAllocRes, finalAllocRes))
+			}
+			zoneName, resName, cmp, ok := cmpAllocatableResources(initialAllocRes, finalAllocRes)
+			framework.Logf("zone=%q resource=%q cmp=%v ok=%v", zoneName, resName, cmp, ok)
+			if !ok {
+				ginkgo.Fail(fmt.Sprintf("failed to compare allocatable resources from node topology initial=%v final=%v", initialAllocRes, finalAllocRes))
+			}
+			// this is actually a workaround. The proper solution is to wait for ALL the container requesting exclusive resources to be gone
```
swatisehgal, line 125: My understanding was that at a given time for a particular test the pod above is running, is that not true? Based on that the `initialAllocRes` and `finalAllocRes` should be the same as the pod acquires CPUs from the shared pool. I am a little confused by this comment here that there might be other exclusive CPUs running?

test/e2e/topology_updater.go
```diff
@@ -41,35 +44,154 @@ import (
...
+		})
+
+		ginkgo.It("it should account for containers requesting exclusive cpus", func() {
+			nodes, err := utils.FilterNodesWithEnoughCores(workerNodes, "1000m")
+			gomega.Expect(err).NotTo(gomega.HaveOccurred())
+			if len(nodes) < 1 {
+				ginkgo.Skip("not enough allocatable cores for this test")
+			}
+
+			ginkgo.By("getting the initial topology information")
```
swatisehgal, line 138: It seems a little unexpected to me that above we are saying that it should account for containers requesting exclusive cpus and then create a pod consuming resources from shared pool, I was expecting a pod requesting exclusive CPU as well. Given that we have a case above where we are only considering a pod in shared pool and resources are not decremented that means that in case we have two pods (one acquiring resources exclusively and another from shared pool)  if resources are decremented in finalAllocRes, it is safe to assume that it is happening because of the pod with exclusive CPUs? WDYT?

swatisehgal (commented):
Thanks for the PR, I took a pass at it. Have a couple of comments/questions.

test/e2e/rte.go
```diff
@@ -182,6 +183,61 @@ var _ = ginkgo.Describe("[RTE][InfraConsuming] Resource topology exporter", func
...
+		ginkgo.It("it should react to pod changes using the smart poller", func() {
+			nodes, err := utils.FilterNodesWithEnoughCores(workerNodes, "1000m")
+			gomega.Expect(err).NotTo(gomega.HaveOccurred())
+			if len(nodes) < 1 {
+				ginkgo.Skip("not enough allocatable cores for this test")
+			}
+
+			initialNodeTopo := getNodeTopology(topologyClient, topologyUpdaterNode.Name, namespace)
+			ginkgo.By("creating a pod consuming the shared pool")
+			sleeperPod := &v1.Pod{
```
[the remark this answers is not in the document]
fromanirh: I guess we can, with properly specialized functions. The intent of inlining the pod definiton is to keep all the context local to the test definition.

test/e2e/rte.go
```diff
@@ -182,6 +183,61 @@ var _ = ginkgo.Describe("[RTE][InfraConsuming] Resource topology exporter", func
...
 
+		ginkgo.It("it should react to pod changes using the smart poller", func() {
+			nodes, err := utils.FilterNodesWithEnoughCores(workerNodes, "1000m")
+			gomega.Expect(err).NotTo(gomega.HaveOccurred())
+			if len(nodes) < 1 {
+				ginkgo.Skip("not enough allocatable cores for this test")
+			}
+
+			initialNodeTopo := getNodeTopology(topologyClient, topologyUpdaterNode.Name, namespace)
+			ginkgo.By("creating a pod consuming the shared pool")
```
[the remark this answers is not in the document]
fromanirh: ack

test/e2e/topology_updater.go
```diff
@@ -41,35 +44,154 @@ import (
...
 		}
 
-		// intentionally get every single time
+		// intentionally done once
 		kubeletConfig, err = e2ekubelet.GetCurrentKubeletConfig(topologyUpdaterNode.Name, "", true)
 		gomega.Expect(err).NotTo(gomega.HaveOccurred())
 	})
 
 	ginkgo.Context("with cluster configured", func() {
+		ginkgo.It("it should not account for any cpus if a container doesn't request exclusive cpus", func() {
```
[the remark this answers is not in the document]
fromanirh: good point. Adding.

test/e2e/topology_updater.go
```diff
@@ -41,35 +44,154 @@ import (
...
+			ginkgo.By("getting the initial topology information")
+			initialNodeTopo := getNodeTopology(topologyClient, topologyUpdaterNode.Name, namespace)
+			ginkgo.By("creating a pod consuming the shared pool")
+			sleeperPod := &v1.Pod{
```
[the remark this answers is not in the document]
fromanirh: a parametric `makePod` function didn't turn to be so helpful in the kube tests; because of the parametric nature the readability ended up a bit worse. But I guess we can try again.

test/e2e/topology_updater.go
fromanirh: The comment is misplaced. Depending on the (random, by design) order on which ginkgo runs the tests, a test which exclusively allocates CPUs may run before. I didn't find yet a clean way to wait for this pod to be completely gone (hence, releasing the exclusively allocated CPUs) before to end the test, so *this* test can run with some leftovers hanging around, which makes the accounting harder.

I will improve and/or move the comment to capture this information.

test/e2e/topology_updater.go
fromanirh: in this case the `ginkgo.By` statement (line 140) is a copy-paste mistake (my fault :) ), this test wants to get exclusively allocated CPUs indeed. Will fix.

test/e2e/utils/pod.go
```diff
@@ -17,37 +17,84 @@ limitations under the License.
...
+			RestartPolicy: v1.RestartPolicyNever,
+			Containers: []v1.Container{
+				v1.Container{
+					Name:  "sleeper-gu-cnt",
+					Image: CentosImage,
+					// 1 hour (or >= 1h in general) is "forever" for our purposes
+					Command: []string{"/bin/sleep", "1h"},
+					Resources: v1.ResourceRequirements{
+						Limits: v1.ResourceList{
+							// we use 1 core because that's the minimal meaningful quantity
```
swatisehgal, line 49: How is this is a guaranteed pod if there is no resource request. For a pod to be guaranteed we have to make sure request=limit ?

test/e2e/topology_updater.go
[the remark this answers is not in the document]
swatisehgal: Oh okay, I saw them being heavily used when I worked on integration tests with SIG scheduling that is why recommended that. Thanks for making the change.

test/e2e/topology_updater.go
swatisehgal: Thanks!

test/e2e/topology_updater.go
```diff
@@ -79,27 +78,52 @@ var _ = ginkgo.Describe("[TopologyUpdater][InfraConsuming] Node topology updater
...
+			if len(initialAllocRes) == 0 || len(finalAllocRes) == 0 {
+				ginkgo.Fail(fmt.Sprintf("failed to find allocatable resources from node topology initial=%v final=%v", initialAllocRes, finalAllocRes))
+			}
+			zoneName, resName, cmp, ok := cmpAllocatableResources(initialAllocRes, finalAllocRes)
+			framework.Logf("zone=%q resource=%q cmp=%v ok=%v", zoneName, resName, cmp, ok)
+			if !ok {
+				ginkgo.Fail(fmt.Sprintf("failed to compare allocatable resources from node topology initial=%v final=%v", initialAllocRes, finalAllocRes))
 			}
 
+			// This is actually a workaround.
```
swatisehgal, line 110: This is definitely a TODO for us but appreciate you providing detailed explanation here.

fromanirh: ack, will mark it as such

test/e2e/utils/pod.go
fromanirh: because https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#requests-and-limits - see note at the bottom of the paragraph.

swatisehgal: Ah nice! Today I learned :) Thanks

PR #10, opened 2021-07-16
gh actions: build the e2e test binary once

fromanirh:
Using the upload/download artifact, we can build the
costly (and big) e2e test binary just once.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -11,10 +11,10 @@ defaults:
     shell: bash
 
 jobs:
-  e2e-ic:
+  e2e-build-artifacts:
+    name: build e2e artifacts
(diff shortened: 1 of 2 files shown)


Tal-or:
Maybe I'm missing something here, but by building the binaries just once, you're missing testing of new changes to those binaries don't you?

fromanirh:
> Maybe I'm missing something here, but by building the binaries just once, you're missing testing of new changes to those binaries don't you?

the idea was to save resources (running time mostly) but this doesn't seem to happen. Two lanes each building their own binaries complete faster (!) than the Y-shaped pipeline we have here.

fromanirh:
nope, the best result is still worse than the previous, so: failed attempt.

PR #12, opened 2021-07-28
make the discovery of the topology manager policy optional

fromanirh:
In this PR we acknowledge the fact that giving access to the kubelet config file to the RTE is sometimes unwanted. Hence
1. we make the kubelet config file path optional and
2. we add an option to just tell the tool the policy, avoiding the need to read the kubelet config file.

### DIFF
--- a/cmd/resource-topology-exporter/main.go
+++ b/cmd/resource-topology-exporter/main.go
@@ -66,12 +66,12 @@ const helpTemplate string = `{{.ProgramName}}
   --export-namespace=<namespace>  Namespace on which update CRDs. Use "" for all namespaces.
   --watch-namespace=<namespace>   Namespace to watch pods for. Use "" for all namespaces.
   --sysfs=<path>                  Top-level component path of sysfs. [Default: /sys]
-  --kubelet-config-file=<path>    Kubelet config file path. [Default: ""]
+  --kubelet-config-file=<path>    Kubelet config file path. [Default: ]
   --topology-manager-policy=<pol> Explicitely set the topology manager policy instead of reading
-                                  from the kubelet. [Default: ""]
+                                  from the kubelet. [Default: ]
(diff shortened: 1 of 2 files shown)


cmd/resource-topology-exporter/main.go
```diff
@@ -65,8 +66,9 @@ const helpTemplate string = `{{.ProgramName}}
   --export-namespace=<namespace>  Namespace on which update CRDs. Use "" for all namespaces.
   --watch-namespace=<namespace>   Namespace to watch pods for. Use "" for all namespaces.
   --sysfs=<path>                  Top-level component path of sysfs. [Default: /sys]
-  --kubelet-config-file=<path>    Kubelet config file path.
-                                  [Default: /kubeletstate/config.yaml]
+  --kubelet-config-file=<path>    Kubelet config file path. [Default: ""]
+  --topology-manager-policy=<pol> Explicitely set the topology manager policy instead of reading
```
swatisehgal, line 70: I feel that we should set the default value here as one of the possible Topology Manager policies. Given that the default TM policy is `none` we should have that as the default value unless overridden by the command line argument or the environment variable. WDYT?

swatisehgal (commented):
Looks good to me, just a small comment about the default value of the Topology Manager policy.

cmd/resource-topology-exporter/main.go
fromanirh: The intention is to error out later in the flow if neither the explicit policy nor the kubelet config file are given:
https://github.com/k8stopologyawareschedwg/resource-topology-exporter/pull/12/commits/2023949196f7f681749568dc1b12a657062f29bb#diff-ecd909a18082b94711e907821b54253df03c1b93cd705230c9cbb2b04e3deb6fR127

swatisehgal: Ack, good enough for now. When we transition to using flags we might be able to do this in a better way but we should keep that in mind.

PR #13, opened 2021-07-29
Improve the e2e tests reliability

fromanirh:
The noderesourcetopology API is being updated, but this project is not ready to consume the new changes yet.
Let's make sure the state is consistent and the tests are reliable.

### DIFF
--- a/hack/check-ds.sh
+++ b/hack/check-ds.sh
@@ -1,0 +1,29 @@
+#!/bin/bash
+
+NAMESPACE="${1:-default}"
+DAEMONSET="${2:-resource-topology-exporter-ds}"
+
+FAILED=1
+
+function wait-for-daemonset(){
+    retries=24
+    while [[ $retries -ge 0 ]];do
+        sleep 5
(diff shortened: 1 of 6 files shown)


hack/check-ds.sh
```diff
@@ -0,0 +1,24 @@
...
+    while [[ $retries -ge 0 ]];do
+        sleep 5
+        ready=$(kubectl -n $NAMESPACE get daemonset $DAEMONSET -o jsonpath="{.status.numberReady}")
+        required=$(kubectl -n $NAMESPACE get daemonset $DAEMONSET -o jsonpath="{.status.desiredNumberScheduled}")
+        if [[ $ready -eq $required ]];then
+            break
+        fi
+        ((retries--))
+        # debug
+        kubectl -n $NAMESPACE describe daemonset $DAEMONSET
```
Tal-or, line 17: Maybe it's better to print it only in case that the ds isn't ready? instead of each time for every iteration

fromanirh: thanks, reworked

PR #14, opened 2021-07-29
podrescli: add option to decorate a client

fromanirh:
This change enables the client wrapping composition at minimal cost.

### DIFF
--- a/pkg/resourcetopologyexporter/resourcetopologyexporter.go
+++ b/pkg/resourcetopologyexporter/resourcetopologyexporter.go
@@ -10,6 +10,8 @@ import (
 
 	"github.com/fsnotify/fsnotify"
 
+	podresourcesapi "k8s.io/kubelet/pkg/apis/podresources/v1"
+
 	"github.com/k8stopologyawareschedwg/resource-topology-exporter/pkg/kubeconf"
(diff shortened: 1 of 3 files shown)


pkg/resourcetopologyexporter/resourcetopologyexporter.go
```diff
@@ -56,19 +58,28 @@ func ContainerIdentFromString(ident string) (*podrescli.ContainerIdent, error) {
 	return cntIdent, nil
 }
 
+func Execute(nrtupdaterArgs nrtupdater.Args, resourcemonitorArgs resourcemonitor.Args, rteArgs Args) error {
```
swatisehgal, line 61: We are no longer using Execute() and are now using the ExecuteWithPodresourcesClient() instead. Has this been left here intentionally or it is a mistake?

swatisehgal (commented):
Looks good to me, just a minor comment.

pkg/resourcetopologyexporter/resourcetopologyexporter.go
fromanirh: It was intentional, but on hindsight it makes no real sense. Let me rename the latter into `Execute`

PR #16, opened 2021-08-05
config: generalize the excludelist configuration

fromanirh:
We generalize the option to read the exclude list from a file
into reading a full configuration file which happens to
contain the exclude list definition.

This change will allow easy expansion of the configuration
at minimal cost.


Tal-or:
/lgtm
Much cleaner!

PR #18, opened 2021-08-09
ci: deflake

fromanirh:
a test started to fail lately; let's fix it.


test/e2e/topology_updater.go
```diff
@@ -69,12 +69,15 @@ var _ = ginkgo.Describe("[TopologyUpdater][InfraConsuming] Node topology updater
 			workerNodes, err = utils.GetWorkerNodes(f)
 			gomega.Expect(err).NotTo(gomega.HaveOccurred())
 
+			// intentionally done once
+			kubeletConfig, err = e2ekubelet.GetCurrentKubeletConfig(nodeName, "", true)
```
Tal-or, line 73: This code will be called once, exactly like before. how it should solve/mitigate the flakiness?

fromanirh: because we don't bail out if this fails. Which is not great, but still we can manage to validate the content of NRTs even without this field.

PR #19, opened 2021-08-09
Prometheus endpoint for RTE

Tal-or:
This update implements a Prometheus endpoint for RTE.
It will expose two different types of metrics:
1. Number of API calls for PodresourceAPI that failed
2. A gauge that represents the latency that took every update of the NRT objects.

The server will listen on port 2112 by default

### DIFF
--- a/cmd/resource-topology-exporter/main.go
+++ b/cmd/resource-topology-exporter/main.go
@@ -10,10 +10,12 @@ import (
 	"time"
 
 	"github.com/docopt/docopt-go"
+
 	"sigs.k8s.io/yaml"
 
 	"github.com/k8stopologyawareschedwg/resource-topology-exporter/pkg/nrtupdater"
 	"github.com/k8stopologyawareschedwg/resource-topology-exporter/pkg/podrescli"
+	"github.com/k8stopologyawareschedwg/resource-topology-exporter/pkg/prometheus"
 	"github.com/k8stopologyawareschedwg/resource-topology-exporter/pkg/resourcemonitor"
 	"github.com/k8stopologyawareschedwg/resource-topology-exporter/pkg/resourcetopologyexporter"
 )
@@ -32,6 +34,11 @@ func main() {
 	cli, err := podrescli.NewFilteringClient(resourcemonitorArgs.PodResourceSocketPath, rteArgs.Debug, rteArgs.ReferenceContainer)
 	if err != nil {
 		log.Fatalf("failed to get podresources client: %v", err)
+	}
+
+	err = prometheus.InitPrometheus()
+	if err != nil {
+		log.Fatalf("failed to start prometheus server: %v", err)
 	}
 
 	err = resourcetopologyexporter.Execute(cli, nrtupdaterArgs, resourcemonitorArgs, rteArgs)
--- a/pkg/nrtupdater/nrtupdater.go
+++ b/pkg/nrtupdater/nrtupdater.go
@@ -7,12 +7,12 @@ import (
 	"os"
 	"time"
 
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
-	v1alpha1 "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/topology/v1alpha1"
-	"k8s.io/apimachinery/pkg/api/errors"
-
+	"github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/topology/v1alpha1"
 	"github.com/k8stopologyawareschedwg/resource-topology-exporter/pkg/dumpobject"
+	"github.com/k8stopologyawareschedwg/resource-topology-exporter/pkg/prometheus"
 )
 
 const (
@@ -126,7 +126,8 @@ func (te *NRTUpdater) Run(infoChannel <-chan MonitorInfo) chan<- struct{} {
 				}
 				tsEnd := time.Now()
 
-				log.Printf("update request received at %v completed in %v", tsBegin, tsEnd.Sub(tsBegin))
+				tsDiff := tsEnd.Sub(tsBegin)
+				prometheus.UpdateOperationDelayMetric("node_resource_object_update", float64(tsDiff.Milliseconds()))
 				if te.args.Oneshot {
 					break
 				}
--- a/pkg/prometheus/prometheus.go
+++ b/pkg/prometheus/prometheus.go
@@ -1,0 +1,85 @@
+package prometheus
+
+import (
+	"fmt"
+	"log"
+	"net/http"
+	"os"
+	"strconv"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+)
+
+const prometheusDefaultPort = "2112"
+
+var nodeName string
+
+var (
+	PodResourceApiCallsFailure = promauto.NewCounterVec(prometheus.CounterOpts{
+		Name: "podresource_api_call_failures_total",
+		Help: "The total number of podresource api calls that failed by the updater",
+	}, []string{"node", "function_name"})
+
+	OperationDelay = promauto.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "operation_delay",
+			Help: "Represent the latency of the update operation",
+		}, []string{"node", "operation_name"})
+)
+
+func getNodeName() (string, error) {
+	var err error
+
+	val, ok := os.LookupEnv("NODE_NAME")
+	if !ok {
+		val, err = os.Hostname()
+		if err != nil {
+			return "", err
+		}
+	}
+	return val, nil
+}
+
+func UpdatePodResourceApiCallsFailureMetric(funcName string) {
+	PodResourceApiCallsFailure.With(prometheus.Labels{
+		"node":      nodeName,
+		"function_name": funcName,
+	}).Inc()
+}
+
+func UpdateOperationDelayMetric(opName string, operationDelay float64) {
+	OperationDelay.With(prometheus.Labels{
+		"node": nodeName,
+		"operation_name": opName,
+	}).Set(operationDelay)
+}
+
+func InitPrometheus() error {
+	var err error
+	var port = prometheusDefaultPort
+
+	if envValue, ok := os.LookupEnv("METRICS_PORT"); ok {
+		if _, err = strconv.Atoi(envValue); err != nil {
+			return fmt.Errorf("the env variable PROMETHEUS_PORT has inccorrect value %q; err %v", envValue, err)
+		}
+		port = envValue
+	}
+
+	nodeName, err = getNodeName()
+	if err != nil {
+		return err
+	}
+
+	http.Handle("/metrics", promhttp.Handler())
+	addr := fmt.Sprintf("0.0.0.0:%s", port)
+
+	go func() {
+		if err = http.ListenAndServe(addr, nil); err != nil {
+			log.Fatalf("failed to run prometheus server; %v", err)
+		}
+	}()
+
+	return nil
+}
--- a/pkg/resourcemonitor/noderesourcesaggregator.go
+++ b/pkg/resourcemonitor/noderesourcesaggregator.go
@@ -34,6 +34,7 @@ import (
 	"github.com/jaypipes/ghw"
 
 	topologyv1alpha1 "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/topology/v1alpha1"
+	"github.com/k8stopologyawareschedwg/resource-topology-exporter/pkg/prometheus"
 )
 
 const (
@@ -69,6 +70,7 @@ func NewResourcesAggregator(sysfsRoot string, podResourceClient podresourcesapi.
 	//Pod Resource API client
 	resp, err := podResourceClient.GetAllocatableResources(ctx, &podresourcesapi.AllocatableResourcesRequest{})
 	if err != nil {
+		prometheus.UpdatePodResourceApiCallsFailureMetric("get_allocatable_resources")
 		return nil, fmt.Errorf("Can't receive response: %v.Get(_) = _, %v", podResourceClient, err)
 	}
 
--- a/pkg/resourcemonitor/podresourcesscanner.go
+++ b/pkg/resourcemonitor/podresourcesscanner.go
@@ -21,8 +21,10 @@ import (
 	"fmt"
 	"log"
 
-	v1 "k8s.io/api/core/v1"
+	"k8s.io/api/core/v1"
 	podresourcesapi "k8s.io/kubelet/pkg/apis/podresources/v1"
+
+	"github.com/k8stopologyawareschedwg/resource-topology-exporter/pkg/prometheus"
 )
 
 type PodResourcesScanner struct {
@@ -61,7 +63,8 @@ func (resMon *PodResourcesScanner) Scan() ([]PodResources, error) {
 	//Pod Resource API client
 	resp, err := resMon.podResourceClient.List(ctx, &podresourcesapi.ListPodResourcesRequest{})
 	if err != nil {
-		return nil, fmt.Errorf("Can't receive response: %v.Get(_) = _, %v", resMon.podResourceClient, err)
+		prometheus.UpdatePodResourceApiCallsFailureMetric("list")
+		return nil, fmt.Errorf("can't receive response: %v.Get(_) = _, %v", resMon.podResourceClient, err)
 	}
 
 	var podResData []PodResources
--- a/pkg/resourcetopologyexporter/resourcetopologyexporter.go
+++ b/pkg/resourcetopologyexporter/resourcetopologyexporter.go
@@ -15,6 +15,7 @@ import (
 	"github.com/k8stopologyawareschedwg/resource-topology-exporter/pkg/kubeconf"
 	"github.com/k8stopologyawareschedwg/resource-topology-exporter/pkg/nrtupdater"
 	"github.com/k8stopologyawareschedwg/resource-topology-exporter/pkg/podrescli"
+	"github.com/k8stopologyawareschedwg/resource-topology-exporter/pkg/prometheus"
 	"github.com/k8stopologyawareschedwg/resource-topology-exporter/pkg/resourcemonitor"
 )
 
@@ -207,8 +208,9 @@ func (rm *ResourceMonitor) Run(eventsChan <-chan PollTrigger) (<-chan nrtupdater
 					Zones: zones,
 				}
 				tsEnd := time.Now()
+				tsDiff := tsEnd.Sub(tsBegin)
 
-				log.Printf("read request received at %v completed in %v", tsBegin, tsEnd.Sub(tsBegin))
+				prometheus.UpdateOperationDelayMetric("podresources_scan", float64(tsDiff.Milliseconds()))
 			case <-done:
 				log.Printf("read stop at %v", time.Now())
 				break
--- a/test/e2e/metrics.go
+++ b/test/e2e/metrics.go
@@ -1,0 +1,53 @@
+package e2e
+
+import (
+	"context"
+
+	"github.com/onsi/ginkgo"
+	"github.com/onsi/gomega"
+
+	"k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/kubernetes/test/e2e/framework"
+)
+
+var _ = ginkgo.Describe("[RTE] metrics", func() {
+	var (
+		initialized bool
+		rtePod      *v1.Pod
+	)
+
+	f := framework.NewDefaultFramework("metrics")
+
+	ginkgo.BeforeEach(func() {
+		if !initialized {
+			pods, err := f.ClientSet.CoreV1().Pods(defaultNamespace).List(context.TODO(), metav1.ListOptions{})
+			gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+			for _, pod := range pods.Items {
+				if val, ok := pod.Labels["name"]; ok && val == rteLabelName {
+					rtePod = &pod
+				}
+			}
+			gomega.Expect(rtePod).ToNot(gomega.BeNil())
+			initialized = true
+		}
+	})
+
+	ginkgo.Context("With prometheus endpoint configured", func() {
+		ginkgo.It("should have some metrics exported", func() {
+			stdout, stderr, err := f.ExecWithOptions(framework.ExecOptions{
+				Command:            []string{"curl", "http://127.0.0.1:2112/metrics"},
+				Namespace:          getNamespaceName(),
+				PodName:            rtePod.Name,
+				ContainerName:      rteContainerName,
+				Stdin:              nil,
+				CaptureStdout:      true,
+				CaptureStderr:      true,
+				PreserveWhitespace: false,
+			})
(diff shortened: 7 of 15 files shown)


fromanirh:
depends on https://github.com/k8stopologyawareschedwg/resource-topology-exporter/pull/18 - let's cut the CI noise.

pkg/nrtupdater/nrtupdater.go
```diff
@@ -126,7 +126,10 @@ func (te *NRTUpdater) Run(infoChannel <-chan MonitorInfo) chan<- struct{} {
 				}
 				tsEnd := time.Now()
 
-				log.Printf("update request received at %v completed in %v", tsBegin, tsEnd.Sub(tsBegin))
+				tsDiff := tsEnd.Sub(tsBegin)
+				prometheus.OperationDelay.Set(float64(tsDiff.Milliseconds()))
+				log.Printf("update request received at %v completed in %v", tsBegin, tsDiff)
```
fromanirh, line 131: please remove this log

pkg/prometheus/prometheus.go
```diff
@@ -0,0 +1,52 @@
...
+var (
+	ApiCallsFailure = promauto.NewCounter(prometheus.CounterOpts{
+		Name: "topology_updater_api_call_failures_total",
+		Help: "The total number of api calls that failed by the updater",
+	})
+
+	OperationDelay = promauto.NewGauge(
+		prometheus.GaugeOpts{
+			Name: "topology_updater_hw_topology_update_operation_measurement",
+			Help: "Represent the latency of the update operation",
```
fromanirh, line 29: ok, the core concept is there, we will need to add labels to distinguish between the operations, and we will probably need one or two more metrics

pkg/resourcemonitor/podresourcesscanner.go
```diff
@@ -61,7 +63,8 @@ func (resMon *PodResourcesScanner) Scan() ([]PodResources, error) {
 	//Pod Resource API client
 	resp, err := resMon.podResourceClient.List(ctx, &podresourcesapi.ListPodResourcesRequest{})
 	if err != nil {
-		return nil, fmt.Errorf("Can't receive response: %v.Get(_) = _, %v", resMon.podResourceClient, err)
+		prometheus.ApiCallsFailure.Inc()
```
fromanirh, line 66: we need to add labels to distinguish the failures, and we need to track failures of `GetAllocatable` as well (even if we know it will fail only because disabled feature gate, basically).

test/e2e/metrics.go
```diff
@@ -0,0 +1,59 @@
...
+import (
+	"context"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/kubernetes/test/e2e/framework"
+
+	topologyclientset "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/generated/clientset/versioned"
+	"github.com/k8stopologyawareschedwg/resource-topology-exporter/test/e2e/utils"
+	"github.com/onsi/ginkgo"
+	"github.com/onsi/gomega"
```
fromanirh, line 12: ginkgo and gomega imports should be moved between `context` and `metav1`:
```
import (
	"context"

        "github.com/onsi/ginkgo"
	"github.com/onsi/gomega"       

	"k8s.io/kubernetes/test/e2e/framework"

        ...
```

test/e2e/metrics.go
```diff
@@ -0,0 +1,59 @@
...
+)
+
+const (
+	rteContainerName = "resource-topology-exporter-container"
+)
```
fromanirh, line 17: minor/nit: this should be probably moved into `e2e_test.go`

test/e2e/metrics.go
```diff
@@ -0,0 +1,59 @@
...
+	f := framework.NewDefaultFramework("metrics")
+
+	ginkgo.BeforeEach(func() {
+		var err error
+
+		if !initialized {
+			_, err = topologyclientset.NewForConfig(f.ClientConfig())
+			gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+			_, err = f.ClientSet.CoreV1().Nodes().Get(context.TODO(), getNodeName(), metav1.GetOptions{})
```
fromanirh, line 32: what is this for, if we throw away the result and keep only the error?

test/e2e/metrics.go
```diff
@@ -0,0 +1,59 @@
...
+				Command:            []string{"curl", "http://127.0.0.1:2112/metrics"},
+				Namespace:          getNamespaceName(),
+				PodName:            pods[0].Name,
+				ContainerName:      rteContainerName,
+				Stdin:              nil,
+				CaptureStdout:      true,
+				CaptureStderr:      true,
+				PreserveWhitespace: false,
+			})
+			gomega.Expect(err).NotTo(gomega.HaveOccurred(), "%v", stderr)
```
fromanirh, line 55: minor/nit: I'd use "%s"

test/e2e/metrics.go
```diff
@@ -0,0 +1,59 @@
...
+			gomega.Expect(stdout).ToNot(gomega.BeEmpty())
```
fromanirh, line 56: Let's look in this output and make sure that the metrics we added are in there. Nothing fancy: `strings.Contains` (or similar) will be fine, no need for regexes.

test/e2e/utils/pod.go
```diff
@@ -98,3 +100,23 @@ func DeletePodSyncByName(f *framework.Framework, podName string) {
...
 	f.PodClient().DeleteSync(podName, delOpts, framework.DefaultPodDeletionTimeout)
 }
+
+func GetByRegex(cs clientset.Interface, reg string) ([]*v1.Pod, error) {
+    podNameRgx, err := regexp.Compile(reg)
+    if err != nil {
+        return nil, err
+    }
+
+    podList, err := cs.CoreV1().Pods("default").List(context.TODO(), metav1.ListOptions{})
```
fromanirh, line 110: let's not hardcode the namespace, better to pass it as argument

test/e2e/metrics.go
Tal-or: Yes, the first one is redundant but the second one will hint if the system is ready for test

cmd/resource-topology-exporter/main.go
```diff
@@ -3,6 +3,7 @@ package main
 import (
 	"bytes"
 	"fmt"
+	"github.com/k8stopologyawareschedwg/resource-topology-exporter/pkg/prometheus"
```
fromanirh, line 6: please order the import like
```
stdlib

utility packages (e.g. yaml, ginkgo...)

kube(-related) packages

local packages
```

cmd/resource-topology-exporter/main.go
```diff
@@ -34,6 +36,13 @@ func main() {
 		log.Fatalf("failed to get podresources client: %v", err)
 	}
 
+	err = prometheus.InitPrometheus()
+	if err != nil {
+		log.Fatalf("failed to deploy prometheus server: %v", err)
+	}else {
+		log.Printf("prometheus server deployed successfully")
```
fromanirh, line 43: this should be a debug log; additionally, I's use "started" rather than "deployed"

fromanirh:
we need to figure out the labeling scheme, I'll do some research

cmd/resource-topology-exporter/main.go
Tal-or: The IDE always messed that up. will fix

Tal-or:
> we need to figure out the labeling scheme, I'll do some research

I didn't change the label yet, I just wanted to make sure the CI tests are fine. i'll look into the labels now

cmd/resource-topology-exporter/main.go
Tal-or: The standard log library doesn't have a debug log by default. What other alternative do we have?

fromanirh: just remove the line then

pkg/prometheus/prometheus.go
```diff
@@ -0,0 +1,52 @@
+package prometheus
+
+import (
+	"fmt"
+	"net/http"
+	"os"
+	"strconv"
+	"time"
+
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
```
fromanirh, line 10: do we actually need these deps? it seems to me that we can just do
```
go func() {
    if err :=  http.ListenAndServe(addr, nil); err != nil {
       log.Fatalf("failed to run prometheus server; %v", err)
   }
}()
```

test/e2e/metrics.go
fromanirh: ok, if you need to wait for the nodes to be ready (do we actually need? can't we just run the metrics scraping in a Eventually block and just wait and bail out on error?), better to do that explicitely.

test/e2e/utils/pod.go
```diff
@@ -17,14 +17,17 @@ limitations under the License.
...
 	"sync"
 
 	"github.com/onsi/ginkgo"
 
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/kubernetes/test/e2e/framework"
+    clientset "k8s.io/client-go/kubernetes"
+    "k8s.io/kubernetes/test/e2e/framework"
```
fromanirh, line 30: please run `gofmt` on sources before uploading. Or, even better, let's add a CI lane to enforce this check :)

test/e2e/metrics.go
```diff
@@ -0,0 +1,55 @@
...
+				PodName:            pods[0].Name,
+				ContainerName:      rteContainerName,
+				Stdin:              nil,
+				CaptureStdout:      true,
+				CaptureStderr:      true,
+				PreserveWhitespace: false,
+			})
+			gomega.Expect(err).NotTo(gomega.HaveOccurred(), "%s", stderr)
+			gomega.Expect(stdout).To(gomega.ContainSubstring("topology_updater_api_call_failures_total"))
+			gomega.Expect(stdout).To(gomega.ContainSubstring("topology_updater_hw_topology_update_operation_measurement"))
```
fromanirh, line 52: now that I see this I think I'll need to review the naming. I'll send some suggestions alongside the labeling suggestions.

pkg/nrtupdater/nrtupdater.go
fromanirh: we need to record the delay also in
```func (rm *ResourceMonitor) Run(eventsChan <-chan PollTrigger) (<-chan nrtupdater.MonitorInfo, chan<- struct{}) {```
in
`pkg/resourcetopologyexporter/resourcetopologyexporter.go`
We will need to use labels to differentiate the two values.

File: test/e2e/metrics.go (hunk 6 above)
Comment (fromanirh) [reply test/e2e/metrics.go:27]: On second thought yes, adding a sync point to make sure the nodes are ready gives us a clearer error message. But let's do explicitely.

Comment (Tal-or) [reply test/e2e/metrics.go:27]: What do you mean by doing that explicitly? something like ```Eventually()```?

Comment (fromanirh) [reply test/e2e/metrics.go:27]: If we need to check the nodes are ready, for example, let's actually make logic on the node objects instead of only checking the error message of the `Get`, which tells us nothing about the state of the nodes.
The code as-is will only tell us that a given node _exists_.

File: pkg/prometheus/prometheus.go [hunk 15]
```diff
@@ -0,0 +1,80 @@
...
+	if strings.Contains(error, "GetAllocatableResources disabled") {
+		return "feature-gate"
+	}
+	return "unexpected"
+}
+
+func nodeName() string {
+	val, ok := os.LookupEnv("NODE_NAME")
+	if !ok {
+		return ""
```
fromanirh, line 42: we need a fallback here. Let's use `https://pkg.go.dev/os#Hostname`. Even better: get the hostname as commandline flag and pass it through. Or get once in `InitPrometheus`

test/e2e/utils/pod.go
```diff
@@ -39,7 +39,7 @@ func MakeGuaranteedSleeperPod(cpuLimit string) *v1.Pod {
 		Spec: v1.PodSpec{
 			RestartPolicy: v1.RestartPolicyNever,
 			Containers: []v1.Container{
-				v1.Container{
+				{
```
fromanirh, line 42: Unnecessary, but no big deal

test/e2e/utils/pod.go
```diff
@@ -66,7 +66,7 @@ func MakeBestEffortSleeperPod() *v1.Pod {
...
```
fromanirh, line 69: same

pkg/nrtupdater/nrtupdater.go
```diff
@@ -126,7 +126,9 @@ func (te *NRTUpdater) Run(infoChannel <-chan MonitorInfo) chan<- struct{} {
 				}
 				tsEnd := time.Now()
 
-				log.Printf("update request received at %v completed in %v", tsBegin, tsEnd.Sub(tsBegin))
+				tsDiff := tsEnd.Sub(tsBegin)
+				prometheus.UpdateOperationDelayMetric(float64(tsDiff.Milliseconds()))
```
fromanirh, line 130: we need to do the same in `pkg/resourcetopologyexporter/resourcetopologyexporter.go:199`:

pkg/resourcemonitor/podresourcesscanner.go
```diff
@@ -61,7 +63,8 @@ func (resMon *PodResourcesScanner) Scan() ([]PodResources, error) {
 	//Pod Resource API client
 	resp, err := resMon.podResourceClient.List(ctx, &podresourcesapi.ListPodResourcesRequest{})
 	if err != nil {
-		return nil, fmt.Errorf("Can't receive response: %v.Get(_) = _, %v", resMon.podResourceClient, err)
+		prometheus.UpdateApiCallsFailureMetric(err)
+		return nil, fmt.Errorf("can't receive response: %v.Get(_) = _, %v", resMon.podResourceClient, err)
```
fromanirh, line 67: we need to record the errors returned by `GetAllocatable` (pkg/resourcemonitor/resourcemonitor.go:188)

pkg/prometheus/prometheus.go
```diff
@@ -0,0 +1,80 @@
...
+	OperationDelay = promauto.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "operation_delay",
+			Help: "Represent the latency of the update operation",
+		}, []string{"node"})
+)
+
+func extractErrorType(error string) string {
+	// TODO temporary until PodresourceAPI errors will be standardize
+	if strings.Contains(error, "GetAllocatableResources disabled") {
```
fromanirh, line 33: I'm always wary when we make logic on the error text. I'd just add a label to track the API being called, and record the raw number of errors. We will differentiate later, if needed.

pkg/prometheus/prometheus.go
```diff
@@ -0,0 +1,80 @@
...
+		return ""
+	}
+	return val
+}
+
+func UpdateApiCallsFailureMetric(err error) {
+	errorType := extractErrorType(err.Error())
+	ApiCallsFailure.With(prometheus.Labels{
+		"node":      nodeName(),
+		"errorType": errorType,
```
fromanirh, line 51: I'm not sure labels use the `camelCase`, let's check the standard.

test/e2e/metrics.go
```diff
@@ -0,0 +1,58 @@
...
+
+	"k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/kubernetes/test/e2e/framework"
+)
+
+var _ = ginkgo.Describe("[RTE] metrics", func() {
+	var (
+		initialized bool
+		rtePod      *v1.Pod = nil
```
fromanirh, line 17: unnecessary initialization, but no big deal

test/e2e/metrics.go
```diff
@@ -0,0 +1,58 @@
...
+	ginkgo.BeforeEach(func() {
+		if !initialized {
+			ds, err := f.ClientSet.AppsV1().DaemonSets(defaultNamespace).Get(context.TODO(), rteDsName, metav1.GetOptions{})
+			gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+			pods, err := f.ClientSet.CoreV1().Pods(defaultNamespace).List(context.TODO(), metav1.ListOptions{})
+			gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+			for _, pod := range pods.Items {
+				for _, or := range pod.GetOwnerReferences() {
```
fromanirh, line 31: or just lookup by a well-known (and documented) label, adding it if needed

pkg/prometheus/prometheus.go
fromanirh: to clarify: the key point is to get the hostname once and pass it through. Taking it from the env var is just fine, actually using a commandline arg is slightly worse option. We need the fallback using `os.Hostname`, though.

test/e2e/utils/pod.go
Tal-or: part of the gofmt run

pkg/prometheus/prometheus.go
Tal-or: We are tracking the number of  API calls using the metric itself (each call increase the metric by 1).
The label is in order to differentiate between feature-gate/non-feature-gate errors.
AFAIU we don't need a label to track the general number of calls because that is to the whole purpose of the metric itself.

fromanirh: then how we can report the `List` errors, which must always be zero (otherwise something horrible is happening)?

pkg/prometheus/prometheus.go
Tal-or: According to standard:
> Prometheus metrics and label names are written in snake_case

Will fix

pkg/prometheus/prometheus.go
Tal-or: Probably I'm missing something.
The ```prometheus.ApiCallsFailure``` should track all types of API failures (any type of call to the API server, ```List```, ```Get``` etc. ) or just the PodreasourceAPI calls?

fromanirh: This metric must count the failures of all the `podresources` API calls, thus `List` and `GetAllocatableResources`. The former count (`List`) must always be zero, and we must report this.

Tal-or: Summarize it here for documentation purposes:
There are two types of podresourceapi calls:
List https://github.com/kubernetes/kubernetes/blob/release-1.22/pkg/kubelet/apis/podresources/server_v1.go#L50 
and GetAllocatableResources https://github.com/kubernetes/kubernetes/blob/release-1.22/pkg/kubelet/apis/podresources/server_v1.go#L82

test/e2e/metrics.go
Tal-or: So add a label in the manifest of the ds? what is the name of the label should look like?

fromanirh: We can add something like
```
  template:
    metadata:
      labels:
        name: resource-topology
        name: topology-updater
```
in https://github.com/k8stopologyawareschedwg/resource-topology-exporter/blob/master/manifests/resource-topology-exporter-ds.yaml#L10

and here in the test just look for a pod labeled `topology-updater` (AND `resource-topology`?). But can be done later, not blocking for this PR.

Tal-or: Looking into it right now. address all the others

pkg/resourcetopologyexporter/resourcetopologyexporter.go
```diff
@@ -207,8 +208,9 @@ func (rm *ResourceMonitor) Run(eventsChan <-chan PollTrigger) (<-chan nrtupdater
 					Zones: zones,
 				}
 				tsEnd := time.Now()
+				tsDiff := tsEnd.Sub(tsBegin)
 
-				log.Printf("read request received at %v completed in %v", tsBegin, tsEnd.Sub(tsBegin))
+				prometheus.UpdateOperationDelayMetric(float64(tsDiff.Milliseconds()))
```
fromanirh, line 213: we need a label to differentiate the updates

pkg/prometheus/prometheus.go
```diff
@@ -0,0 +1,84 @@
...
+
+const prometheusDefaultPort = "2112"
+
+var nodeName string
+
+var (
+	PodResourceApiCallsFailure = promauto.NewCounterVec(prometheus.CounterOpts{
+		Name: "podresource_api_call_failures_total",
+		Help: "The total number of podresource api calls that failed by the updater",
+	}, []string{"node", "call_type"})
```
fromanirh, line 23: I'd just use "name" or "function_name", this doesn't really seem a "call type".

test/e2e/metrics.go
Tal-or: We can't have the same key twice, but anyway I see there is only one pod (per node) with the label ```name=resource-topology``` so this is enough I think

pkg/nrtupdater/nrtupdater.go
```diff
@@ -126,7 +126,8 @@ func (te *NRTUpdater) Run(infoChannel <-chan MonitorInfo) chan<- struct{} {
 				}
 				tsEnd := time.Now()
 
-				log.Printf("update request received at %v completed in %v", tsBegin, tsEnd.Sub(tsBegin))
+				tsDiff := tsEnd.Sub(tsBegin)
+				prometheus.UpdateOperationDelayMetric("nrtUpdate", float64(tsDiff.Milliseconds()))
```
fromanirh, line 130: please replace "nrtUpdate" with `node_resource_object_update`

pkg/resourcemonitor/noderesourcesaggregator.go
```diff
@@ -69,6 +70,7 @@ func NewResourcesAggregator(sysfsRoot string, podResourceClient podresourcesapi.
 	//Pod Resource API client
 	resp, err := podResourceClient.GetAllocatableResources(ctx, &podresourcesapi.AllocatableResourcesRequest{})
 	if err != nil {
+		prometheus.UpdatePodResourceApiCallsFailureMetric("GetAllocatableResources")
```
fromanirh, line 73: I'm not sure if the values are expected to be `snake_case` or not.

pkg/resourcetopologyexporter/resourcetopologyexporter.go
```diff
@@ -207,8 +208,9 @@ func (rm *ResourceMonitor) Run(eventsChan <-chan PollTrigger) (<-chan nrtupdater
 					Zones: zones,
 				}
 				tsEnd := time.Now()
+				tsDiff := tsEnd.Sub(tsBegin)
 
-				log.Printf("read request received at %v completed in %v", tsBegin, tsEnd.Sub(tsBegin))
+				prometheus.UpdateOperationDelayMetric("podResScan", float64(tsDiff.Milliseconds()))
```
fromanirh, line 213: please replace "podResScan" with `podresources_scan`

fromanirh: we should probably add another label with the content of `pt.Timer` here.

fromanirh (commented):
Almost there. Minor fixes to the labels and we are done.

pkg/resourcetopologyexporter/resourcetopologyexporter.go
fromanirh: Actually, the value of the label should be the return value of https://github.com/k8stopologyawareschedwg/resource-topology-exporter/blob/master/pkg/nrtupdater/nrtupdater.go#L58

fromanirh: on second thought, let's postpone this change

fromanirh (commented):
/approve
/lgtm

PR #21, opened 2021-08-17
prometheus: add wakeup delay metrics

fromanirh:
When reporting the operation delay metric, it's useful
to track which reason we had to trigger an operation.

For node resource objects, it's easy: it's always reactive, so we
hardcode the label; for podresources scan, it could be either
reactive (smart polling) or periodic (trigger update)

### DIFF
--- a/test/e2e/metrics.go
+++ b/test/e2e/metrics.go
@@ -2,6 +2,8 @@ package e2e
 
 import (
 	"context"
+	"fmt"
+	"strconv"
 
 	"github.com/onsi/ginkgo"
 	"github.com/onsi/gomega"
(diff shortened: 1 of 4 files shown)


test/e2e/metrics.go
```diff
@@ -2,11 +2,13 @@ package e2e
...
 import (
 	"context"
+	"fmt"
+	"strconv"
 
 	"github.com/onsi/ginkgo"
 	"github.com/onsi/gomega"
 
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
fromanirh, line 11: should be `corev1`

PR #22, opened 2021-08-18
gh actions: add release flow

fromanirh:
Add github action to release when a new tag is created.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,130 @@
+name: Release
+
+on:
+  push:
+    # Sequence of patterns matched against refs/tags
+    tags:
+      - "v[0-9]+.[0-9]+.[0-9]+"
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  release-build:
+    runs-on: ubuntu-20.04
+    env:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,130 @@
...
+        go-version: 1.16
+
+    - name: verify modules
+      run: go mod verify
+
+    - name: set release version env var
+      run: |
+        echo "RELEASE_VERSION=${GITHUB_REF#refs/*/}" >> $GITHUB_ENV
+
+    - name: build test binary
```
Tal-or, line 38: Here the execution code is the same as the gh ```e2e.yml``` 
Maybe we can write the ci test as a separate action and then we can reuse it like:
```
- name: test
  uses: ./.github/actions/ci-e2e
 ```
 But this can be done on a separate PR

fromanirh: this is a neat idea; my only concern is: the above is a workflow (snippet), not a proper action. I'm not sure how easy it is to wrap up these steps in a proper action.

Tal-or: Let me do some homework and see

PR #23, opened 2021-08-18
Remove docopt library

Tal-or:
Replace docopt with the std go flag library.
In addition, added unit tests to make sure flags are parsed properly.

Still need to tail the unit tests in such way that it will run automatically by the CI


cmd/resource-topology-exporter/main.go
```diff
@@ -25,13 +23,25 @@ const (
 	ProgramName = "resource-topology-exporter"
 )
 
+type progArgs struct {
+	nrtupdater      *nrtupdater.Args
+	resourcemonitor *resourcemonitor.Args
+	rte             *resourcetopologyexporter.Args
+	version         *bool
```
fromanirh, line 30: why do we need pointers here?

cmd/resource-topology-exporter/main.go
```diff
@@ -25,13 +23,25 @@ const (
...
+
 func main() {
-	nrtupdaterArgs, resourcemonitorArgs, rteArgs, err := argsParse(os.Args[1:])
+	pArgs, err := parseArgs(os.Args[1:]...)
 	if err != nil {
-		log.Fatalf("failed to parse command line: %v", err)
+		log.Fatalf("failed to parse args: %v", err)
+	}
+
+	if *pArgs.version {
```
fromanirh, line 39: I don't think you need to deference `pArgs` anyway

cmd/resource-topology-exporter/main.go
```diff
@@ -41,176 +51,103 @@ func main() {
...
 	}
-	err = tmpl.Execute(&helpBuffer, helpData)
-	if err != nil {
-		return "", err
+	err = yaml.Unmarshal(data, &conf)
+	return conf, err
+}
+
+// The args is passed only for testing purposes.
+func parseArgs(args ...string) (*progArgs, error) {
```
fromanirh, line 80: why do we need to return a pointer to `progArgs`?

cmd/resource-topology-exporter/main.go
```diff
@@ -41,176 +51,103 @@ func main() {
...
-	}
-	resourcemonitorArgs.SysfsRoot = arguments["--sysfs"].(string)
-	if path, ok := arguments["--podresources-socket"].(string); ok {
-		resourcemonitorArgs.PodResourceSocketPath = path
+		return pArgs, err
 	}
 
-	if kubeletStateDirs, ok := arguments["--kubelet-state-dir"].([]string); ok {
-		resourcemonitorArgs.KubeletStateDirs = kubeletStateDirs
+	if reflect.DeepEqual(*pArgs.rte.ReferenceContainer, podrescli.ContainerIdent{}) {
```
fromanirh, line 127: just check if the contents are non-zero value and/or add a helper method into `podrescli.ContainerIdent`. maybe `IsEmpty`.

cmd/resource-topology-exporter/main.go
```diff
@@ -41,176 +51,103 @@ func main() {
...
-	conf := config{}
-	data, err := os.ReadFile(configPath)
-	if err != nil {
-		// config is optional
-		if errors.Is(err, os.ErrNotExist) {
-			log.Printf("Info: couldn't find configuration in %q", configPath)
-			return conf, nil
-		}
-		return conf, err
+func defaultTmPolicy() string {
```
fromanirh, line 147: I think here we can afford to write `defaultTopologyManagerPolicy`

cmd/resource-topology-exporter/main_suite_test.go
```diff
@@ -0,0 +1,28 @@
+package main
+
+import (
+	"os"
+	"testing"
+
+	. "github.com/onsi/ginkgo"
```
fromanirh, line 7: I'm not that fond of ginkgo for unit tests (see: pkg/kubeconf/kubelet_config_file_test.go). We use in general `convey` and/or `ginkgo` only for code which we want to merge in other projects (NFD) which already use this infra.
Or, from the other perspective, what do we gain in using ginkgo in unit tests?

pkg/podrescli/client.go
```diff
@@ -33,6 +35,48 @@ func (ci *ContainerIdent) String() string {
 	return fmt.Sprintf("%s/%s/%s", ci.Namespace, ci.PodName, ci.ContainerName)
 }
 
+func (ci *ContainerIdent) Set(val string) error {
```
fromanirh, line 38: not a fan of these interface leaking into utility packages. Do we have other options?

pkg/podrescli/client.go
```diff
@@ -33,6 +35,48 @@ func (ci *ContainerIdent) String() string {
...
+func ContainerIdentFromEnv() *ContainerIdent {
+	cntIdent := ContainerIdent{
+		Namespace:     os.Getenv("REFERENCE_NAMESPACE"),
+		PodName:       os.Getenv("REFERENCE_POD_NAME"),
+		ContainerName: os.Getenv("REFERENCE_CONTAINER_NAME"),
+	}
+	if cntIdent.Namespace == "" || cntIdent.PodName == "" || cntIdent.ContainerName == "" {
+		return nil
+	}
+	return &cntIdent
```
fromanirh, line 60: we can extract this code and add a
```
func (ci *ContainerIdent) IsEmpty() bool {
    return cntIdent.Namespace == "" || cntIdent.PodName == "" || cntIdent.ContainerName == ""
}
```

pkg/resourcemonitor/types.go
```diff
@@ -30,13 +32,26 @@ type Args struct {
...
 	SysfsRoot             string
 	KubeletConfigFile     string
-	KubeletStateDirs      []string
+	KubeletStateDirs      KubeletStateDirs
 	ExcludeList           ResourceExcludeList
 }
 
+type KubeletStateDirs []string
+
+func (ksd *KubeletStateDirs) String() string {
```
fromanirh, line 41: again, not really a fan of these interface leaking into utility packages.

cmd/resource-topology-exporter/main_test.go
```diff
@@ -0,0 +1,53 @@
...
+		})
+
+		It("must parse custom flags properly", func() {
+			pArgs, err = parseArgs("--kubelet-state-dir=dir1 dir2 dir3", "--reference-container=ns/pod/cont")
+			Expect(err).ToNot(HaveOccurred())
+			Expect(pArgs.resourcemonitor.KubeletStateDirs).To(BeEquivalentTo([]string{"dir1", "dir2", "dir3"}))
+			Expect(pArgs.rte.ReferenceContainer).To(Equal(&podrescli.ContainerIdent{Namespace: "ns", PodName: "pod", ContainerName: "cont"}))
+		})
+
+		It("should have the following default values", func() {
```
fromanirh, line 36: mixed feelings here. We should probably keep this test, but only after having added a way to initialize all the structs 
with default values (without, somehow, duplicating the values we set from flags), and use something the golden file pattern (see: https://about.sourcegraph.com/go/advanced-testing-in-go/#Golden-files-also-test-flags)

IOW: the default values should be taken from a file, not buried in code.

cmd/resource-topology-exporter/main.go
Tal-or: Because ```flags.BoolVar()``` get a pointer and the sole purpose of the struct (which hold ```version```) is to use those ```flags.TypeVar()``` functions so it's more natural to pass ```version``` and not ```&version```

cmd/resource-topology-exporter/main.go
Tal-or: Right. My C++ nature

cmd/resource-topology-exporter/main.go
Tal-or: Avoid redundant copy.

cmd/resource-topology-exporter/main.go
fromanirh: I don't fully agree about version, but what about the other `Args`? Why do we need pointers to them?

cmd/resource-topology-exporter/main.go
fromanirh: Which happens once however. Do we have stronger reasons?
http://macias.info/entry/201802102230_go_values_vs_references.md
https://medium.com/@philpearl/bad-go-pointer-returns-340f2da8289

cmd/resource-topology-exporter/main.go
Tal-or: This will allow us to expand ```ReferenceContainer``` without worries about maintaining the ```IsEmpty``` or adding additional comparisons don't you think?

cmd/resource-topology-exporter/main.go
Tal-or: Sure

cmd/resource-topology-exporter/main.go
fromanirh: ok, it's a bit redundant and non-idiomatic, I'd say:
https://golang.org/ref/spec#Selectors

cmd/resource-topology-exporter/main.go
fromanirh: I strongly prefer an explicit method (one-liner anyway) unless there is another reason I don't see atm. The method itself can use `reflect.DeepEqual` internally if needed (not a fan of it but I'm open to try)

cmd/resource-topology-exporter/main_suite_test.go
Tal-or: 1. I used PAO project as a reference and I saw we're using ginkgo for unit tests there as well.
2. Consistency - Use the same framework for both e2e & unit test
3. Nit: I'm more experience with ginkgo :)

pkg/podrescli/client.go
Tal-or: Yes we do:
1. Use additional helper string variable and drop this ```Set``` function
2. move the logic into somewhere else?
BTW what is the reason that you don't like interface leaking into utility packages?

pkg/podrescli/client.go
Tal-or: sure

cmd/resource-topology-exporter/main_suite_test.go
fromanirh: thing is, different tests serve different purposes. So using the same framework for both is actually a double edged sword. Since we don't really have consistency in this project, I don't really mind, but I'm (strongly) inclined to use the plain `testing` module for unit tests, which is usually more than up to the task, unless there are clear reasons to use something more featureful.

cmd/resource-topology-exporter/main_test.go
Tal-or: Cool I'll try that

pkg/podrescli/client.go
fromanirh: > BTW what is the reason that you don't like interface leaking into utility packages?
Separation of concerns, minimization of inter-package deps/more loose coupling.
In this specific case I don't see (yet) many gains about using an intermediate variable, but we add a dep between packages. Doesn't seem an obvious gain to me.

cmd/resource-topology-exporter/main.go
Tal-or: Yes as I said, need to check. np I'll change that.

cmd/resource-topology-exporter/main.go
Tal-or: Cool articles, wasn't aware of that. Will fix

pkg/podrescli/client.go
Tal-or: > In this specific case I don't see (yet) many gains about using an intermediate variable

In the ```main.go``` we can do something like:
```
flags.StringVar(&contRef, "container-reference", "", "usage description") 
ci := ContainerIdentFromString(contRef)
```
and we don't need the ```Set``` function anymore, so you eliminate the interface leaking into utility package

cmd/resource-topology-exporter/main.go
Tal-or: Now that I read this: https://github.com/k8stopologyawareschedwg/resource-topology-exporter/pull/23#discussion_r691819551
I'll replace that.

cmd/resource-topology-exporter/main.go
Tal-or: hmm I checked and I can't refer to the ```version``` without dereferencing ```pArgs``` first

cmd/resource-topology-exporter/main.go
fromanirh: sure, my point here is that using pointer values instead of copying values is not obviously better every time, and it has its own fair share of drawbacks. So it should be a conscious decision when do what. I for myself I believe there is unwarranted fear about returning and passing by "large" values. We should not just default to "use pointer, it's faster".

pkg/podrescli/client.go
fromanirh: that seems fine.

Tal-or:
Depends on https://github.com/k8stopologyawareschedwg/resource-topology-exporter/pull/24

cmd/resource-topology-exporter/main.go
```diff
@@ -25,13 +24,29 @@ const (
...
-	nrtupdaterArgs, resourcemonitorArgs, rteArgs, err := argsParse(os.Args[1:])
+	parsedArgs, err := parseArgs(os.Args[1:]...)
 	if err != nil {
-		log.Fatalf("failed to parse command line: %v", err)
+		log.Fatalf("failed to parse args: %v", err)
 	}
 
-	cli, err := podrescli.NewFilteringClient(rteArgs.PodResourcesSocketPath, rteArgs.Debug, rteArgs.ReferenceContainer)
+	if parsedArgs.Version {
+		fmt.Println(ProgramName, "TODO add Version track logic")
```
fromanirh, line 45: do we have plans to address this?

cmd/resource-topology-exporter/main.go
```diff
@@ -41,179 +56,131 @@ func main() {
...
+
+func defaultHostName() string {
+	var err error
+
+	val, ok := os.LookupEnv("NODE_NAME")
+	if !ok || val == "" {
+		val, err = os.Hostname()
 		if err != nil {
-			log.Fatalf("error getting exclude list from the configuration: %v", err)
+			log.Fatalf("error getting the host name: %v", err)
```
fromanirh, line 153: I'm not sure we should `Fatalf` here, but we can improve later

fromanirh (approved):
/lgtm

cmd/resource-topology-exporter/main.go
Tal-or: Yes, we do. would you like me to issue a card?

fromanirh: If  you can file a PR today, no need for a card; otherwise yes, please do.

Tal-or: Will file a PR today

PR #24, opened 2021-08-23
Unittests: Sort []ContainerDevices before comparison

Tal-or:
In some case like can be shown here: https://github.com/k8stopologyawareschedwg/resource-topology-exporter/pull/23/checks?check_run_id=3393185481#step:5:23
the test might failed due to different order of objects between the slices.


fromanirh (approved):
/lgtm

PR #27, opened 2021-08-23
Update golden files before running test-unit target

Tal-or:
This will assure testing the default value for any new flag that will be added


Makefile
```diff
@@ -55,7 +55,7 @@ push: image
 	$(RUNTIME) push $(RTE_CONTAINER_IMAGE)
 
 .PHONY: test-unit
-test-unit:
+test-unit: update-golden-files
```
fromanirh, line 58: Could you please add a bit more context (in the commit message)? What's the flow you are automating here?
Every time I see a test which tries to auto-fix itself (to some extent) I become a bit wary

Tal-or: We have the [TestArgsParse.default.json](https://github.com/k8stopologyawareschedwg/resource-topology-exporter/blob/master/test/data/TestArgsParse.default.json) AKA golden file, that hold the default values of all flags and the output of the ```argsParse()``` compared against this file during the UT.

Assuming a developer is adding a new flag, the ```TestArgsParse.default.json``` content is not valid anymore and we should update it. 

This is done using this new make target.
If you think it's too much to run this for every ```make test-unit``` we can just keep the make target and run it manually

PR #28, opened 2021-08-24
Add versioning mechanism

Tal-or:
1. The version will be defined in build time using ldflags
2. Added an e2e to make sure the version being injected properly

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,12 +9,19 @@ IMAGENAME ?= resource-topology-exporter
 IMAGETAG ?= latest
 RTE_CONTAINER_IMAGE ?= quay.io/$(REPOOWNER)/$(IMAGENAME):$(IMAGETAG)
 
+VERSION := $(shell git tag --sort=committerdate | head -n 1)
+ifeq ($(VERSION),)
+	VERSION = v0.1
+endif
+
+LDFLAGS = -ldflags "-s -w -X github.com/k8stopologyawareschedwg/resource-topology-exporter/pkg/version.version=$(VERSION)"
+
 .PHONY: all
 all: build
 
 .PHONY: build
 build: outdir
-	$(COMMONENVVAR) $(BUILDENVVAR) go build -ldflags '-w' -o _out/resource-topology-exporter cmd/resource-topology-exporter/main.go
+	$(COMMONENVVAR) $(BUILDENVVAR) go build $(LDFLAGS) -o _out/resource-topology-exporter cmd/resource-topology-exporter/main.go
 
 .PHONY: gofmt
 gofmt:
--- a/pkg/version/version.go
+++ b/pkg/version/version.go
@@ -1,0 +1,34 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+Licensed under the Apache License, version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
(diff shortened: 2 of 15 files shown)


test/e2e/rte.go
```diff
@@ -104,6 +107,25 @@ var _ = ginkgo.Describe("[RTE][InfraConsuming] Resource topology exporter", func
...
+			})
+			gomega.Expect(err).ToNot(gomega.HaveOccurred())
+
+			out2, err := execCmd([]string{
+				"git",
+				"describe", "--tags", "--dirty", "--always",
+			})
+			gomega.Expect(err).ToNot(gomega.HaveOccurred())
+
+			gomega.Expect(out1).To(gomega.Equal(out2))
```
fromanirh, line 125: In general, test should not be aware of the implementation, as much as possible. In this case, I'd just check the output of the version is a valid version, possibly using the semantic versioning.

test/e2e/rte.go
```diff
@@ -171,3 +193,12 @@ func cmpResourceList(expected, got v1.ResourceList) (string, int, bool) {
 	}
 	return "", 0, true
 }
+
+func execCmd(cmdline []string) ([]byte, error) {
```
fromanirh, line 197: if we change the test to check the output is a valid version, this can and should be inlined in the calling site.

test/e2e/rte.go
Tal-or: but call ```version.Get()``` from the test will return an undefined version since it's injected only at build time.

fromanirh: If it returns undefined, how it can be equal to the output of `git describe ... `?

fromanirh: Anyway, something along these lines:
```
import (
        ...
	"github.com/hashicorp/go-version"
        ...
)

var _ = ginkgo.Describe("rte version", func() {
	ginkgo.Context("with the tool available", func() {
		ginkgo.It("it should show the correct version", func() {
			cmdline := []string{
				filepath.Join(binariesPath, "resource-topology-exporter"),
				"version",
			}
			fmt.Fprintf(ginkgo.GinkgoWriter, "running: %v\n", cmdline)

			cmd := exec.Command(cmdline[0], cmdline[1:]...)
			cmd.Stderr = ginkgo.GinkgoWriter

			out, err := cmd.Output()
			gomega.Expect(err).ToNot(gomega.HaveOccurred())

			text := strings.TrimSpace(string(out))
			fmt.Fprintf(ginkgo.GinkgoWriter, "reported version: %q\n", text)
			gomega.Expect(text).ToNot(gomega.BeEmpty())
			_, err = version.NewVersion(text)
			gomega.Expect(err).ToNot(gomega.HaveOccurred())
		})
	})
})
```

Tal-or: > If it returns undefined, how it can be equal to the output of `git describe ... `?

Because we're not calling ```version.Get()``` from the code directly, but invoking the binary itself which has the version already injected

Makefile
```diff
@@ -9,12 +9,19 @@ IMAGENAME ?= resource-topology-exporter
 IMAGETAG ?= latest
 RTE_CONTAINER_IMAGE ?= quay.io/$(REPOOWNER)/$(IMAGENAME):$(IMAGETAG)
 
+VERSION := $(shell git tag --sort=committerdate | head -n 1)
+ifeq ($(VERSION),)
+	VERSION = v0.1
```
fromanirh, line 14: in case we can't detect the version from the git history, better to use something clearly identifiable like `v0.0.0` or `-devel`. But we can improve later.

pkg/version/version.go
```diff
@@ -0,0 +1,34 @@
...
+limitations under the License.
+*/
+
+package version
+
+
+const (
+	// ProgramName is the canonical name of this program
+    ProgramName = "resource-topology-exporter"
+	undefinedVersion string = "undefined"
```
fromanirh, line 20: bad formatting, but we can fix later.

fromanirh (commented):
/lgtm

PR #29, opened 2021-08-24
Improve sorting before comparing

Tal-or:
From time to time the resourcemonitor UT failed becasue the structs are equals but in different order.
Usually this is happening because the sorting of the structs isn't good enough.
In this update we added additional sort in which even if the ResourceName is identical we'll sort by DeviceIds, thus it increase the chances for both structs to look exactly the same.
If the error will still occur we'll add sort by third field and so on.

### DIFF
--- a/pkg/resourcemonitor/resourcemonitor_test.go
+++ b/pkg/resourcemonitor/resourcemonitor_test.go
@@ -20,6 +20,7 @@ import (
 	"encoding/json"
 	"log"
 	"sort"
+	"strings"
 	"testing"
 
 	"k8s.io/apimachinery/pkg/util/intstr"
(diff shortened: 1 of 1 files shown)


pkg/resourcemonitor/resourcemonitor_test.go
```diff
@@ -316,10 +317,33 @@ func TestNormalizeContainerDevices(t *testing.T) {
...
+			if res[i].ResourceName == res[j].ResourceName {
+				var sbi, sbj strings.Builder
+				for _, id := range res[i].DeviceIds {
+					sbi.WriteString(id)
+				}
+
+				for _, id := range res[i].DeviceIds {
+					sbj.WriteString(id)
+				}
+				return sbi.String() < sbj.String()
```
fromanirh, line 329: I don't mind too much because it's test code, but other alternatives seen in the wild:
1. deep sort all the nested fields ahead of time
2. compare a pretty print of the nested data structure; of course the pretty printer must ensure traversal of the objects in the correct, stable order (e.g. go-spew does this).

Tal-or: Nice tricks. In case we'll bump into this issue again I'll try that

PR #30, opened 2021-08-26
log: move to klog/v2

fromanirh:
We need a bit more powerful log package to be up to date with
best practices (most notably NFD) and to be able to toggle
messages depending on the verbosiness level.

klog is the default choice in the kube ecosystem, so we switch to it.

### DIFF
--- a/pkg/podrescli/client.go
+++ b/pkg/podrescli/client.go
@@ -3,7 +3,6 @@ package podrescli
 import (
 	"context"
 	"fmt"
-	"log"
 	"os"
 	"reflect"
 	"strings"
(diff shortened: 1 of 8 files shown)


pkg/resourcemonitor/resourcemonitor.go
```diff
@@ -32,6 +31,7 @@ import (
 	podresourcesapi "k8s.io/kubelet/pkg/apis/podresources/v1"
 
 	"github.com/jaypipes/ghw"
+	"k8s.io/klog/v2"
```
Tal-or, line 34: This line should be together with the other k8s dependencies

pkg/podrescli/client.go
```diff
@@ -3,13 +3,13 @@ package podrescli
...
 	"context"
 	"fmt"
-	"log"
 	"os"
 	"reflect"
 	"strings"
 	"time"
 
 	"google.golang.org/grpc"
+	"k8s.io/klog/v2"
```
Tal-or, line 12: This line should be together with the other k8s dependencies

pkg/resourcemonitor/resourcemonitor.go
fromanirh: right, fixed.

pkg/podrescli/client.go
fromanirh: right, fixed.

PR #32, opened 2021-09-14
Sync RTE with NFD

Tal-or:
- Sync podres package
- Bump NRT API to latest:
 - Update vendor files
 - Fix code to match the new API
- Populate Capacity, Available and Allocatable as NFD is doing

TODO (on separate PR) when https://github.com/openshift-kni/resource-topology-exporter/pull/23 and https://github.com/jaypipes/ghw/pull/268 get merged we should take care of memory capacity accounting as well


swatisehgal:
> The available accounting will be added on a separate commit

pkg/resourcemonitor/resourcemonitor_test.go
```diff
@@ -691,23 +691,23 @@ func TestResourcesScan(t *testing.T) {
 					Resources: topologyv1alpha1.ResourceInfoList{
 						topologyv1alpha1.ResourceInfo{
 							Name:        "cpu",
-							Allocatable: intstr.FromString("10"),
-							Capacity:    intstr.FromString("12"),
+							Allocatable: resource.MustParse("10"),
```
swatisehgal, line 694: Wanted to point out how I went ahead with this change in NFD, I assumed that capacity=allocatable (i.e. we don't account for kube-reserved and system reserved) and changed the values in the `allocatable` field to be captured in the `available` field some like I did here: https://github.com/kubernetes-sigs/node-feature-discovery/pull/525/commits/041aa80b31a57dab0832d10081ef19a3ffd1ef9e and then later in a subsequent commit took the reserved resources into consideration: https://github.com/kubernetes-sigs/node-feature-discovery/pull/525/commits/041aa80b31a57dab0832d10081ef19a3ffd1ef9e.

Tal-or:
> > The available accounting will be added on a separate commit
> 
> Just an FYI, the scheduler plugin uses `available` field to make its NUMA aware scheduling decision, so we need to make sure that available field is populated to make sure that it works with the latest version of the scheduler plugin.

This is going to be part of this PR but in a separate commit for readability porpuses

Tal-or:
@swatisehgal @fromanirh FYI the ```Allocatable``` field in RTE is behaving as the ```Available``` filed in NFD
I just tested it on my setup, one you deploy/delete a guaranteed pod, the values under ```Allocatable``` are changing.

https://github.com/k8stopologyawareschedwg/resource-topology-exporter/blob/master/pkg/resourcemonitor/resourcemonitor.go#L171 
is it intentional?

pkg/resourcemonitor/resourcemonitor_test.go
```diff
@@ -18,12 +18,12 @@ package resourcemonitor
 
 import (
 	"encoding/json"
+	"k8s.io/apimachinery/pkg/api/resource"
```
fromanirh, line 21: please put the import in the group below alongside with the other k8s imports

fromanirh:
> @swatisehgal @fromanirh FYI the `Allocatable` field in RTE is behaving as the `Available` filed in NFD.
> I just tested it on my setup, one you deploy/delete a guaranteed pod, the values under `Allocatable` are changing, while it shouldn't, becuase it should reflect the amount of capacity - reserved and it doesn't depend on the number of running pods in the cluster.
> As you can see: https://github.com/k8stopologyawareschedwg/resource-topology-exporter/blob/master/pkg/resourcemonitor/resourcemonitor.go#L171
> is this change in the behavior is intentional?

With nodresourcetopology API 0.0.8 "Allocatable" behaves like "Available" in API 0.0.10:
```
v0.0.8           v0.0.10
N/A               Capacity
Capacity      Allocatable
Allocatable   Available
```

Tal-or:
> With nodresourcetopology API 0.0.8 "Allocatable" behaves like "Available" in API 0.0.10:
> 
> ```
> v0.0.8           v0.0.10
> N/A               Capacity
> Capacity      Allocatable
> Allocatable   Available
> ```

Oh I see. Sorry I wasn't aware of that.

pkg/resourcemonitor/resourcemonitor_test.go
```diff
@@ -474,13 +474,15 @@ func TestResourcesScan(t *testing.T) {
...
-							Capacity:    intstr.FromString("12"),
+							Available:   resource.MustParse("12"),
+							Allocatable: resource.MustParse("12"),
+							Capacity:    resource.MustParse("12"),
 						},
 						topologyv1alpha1.ResourceInfo{
 							Name:        "fake.io/net",
-							Allocatable: intstr.FromString("4"),
-							Capacity:    intstr.FromString("4"),
+							Available:   resource.MustParse("12"),
```
swatisehgal, line 483: This looks like a mistake to me: available should be equal to `4` here as well. Also note that available will always be <= both allocatable and capacity.

Tal-or: yes. copy paste mistake I just found out myself.

fromanirh:
the podres sync patch can be a separate PR which we can merge quickly.

fromanirh:
good enough, we can improve in followup PRs.

PR #34, opened 2021-09-15
NFD resync cleanup

fromanirh:
Cleanup and further refinements following up https://github.com/k8stopologyawareschedwg/resource-topology-exporter/pull/32. All the changes here are expected to preserve the behaviour introduced in the aforementioned PR, aiming to only simplifying the implementation, and expand the test coverage.

### DIFF
--- a/pkg/resourcemonitor/resourcemonitor.go
+++ b/pkg/resourcemonitor/resourcemonitor.go
@@ -89,12 +89,12 @@ func (rc *resourceCounter) HasCPU() bool {
 type perNUMAResourceCounter map[int]resourceCounter
 
 type resourceMonitor struct {
-	nodeName             string
-	args                 Args
-	podResCli            podresourcesapi.PodResourcesListerClient
-	topo                 *ghw.TopologyInfo
-	coreIDToNodeIDMap    map[int]int
-	nodeAllocatable      perNUMAResourceCounter
+	nodeName          string
+	args              Args
+	podResCli         podresourcesapi.PodResourcesListerClient
+	topo              *ghw.TopologyInfo
+	coreIDToNodeIDMap map[int]int
+	nodeAllocatable   perNUMAResourceCounter
 }
 
 func NewResourceMonitor(podResCli podresourcesapi.PodResourcesListerClient, args Args) (*resourceMonitor, error) {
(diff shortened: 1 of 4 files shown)


fromanirh:
@swatisehgal please wait for the WIP to be lifted before to review this PR

swatisehgal:
/lgtm
/approve

Thanks for the PR!

pkg/resourcemonitor/resourcemonitor.go
```diff
@@ -171,51 +162,45 @@ func (rm *resourceMonitor) Scan(excludeList ResourceExcludeList) (topologyv1alph
 
 		// check if NUMA has some allocatable resources
 		resCounters, ok := rm.nodeAllocatable[nodeID]
-		if ok {
-			for resName, resAlloc := range resCounters {
-				if inExcludeSet(excludeSet, resName, rm.nodeName) {
-					continue
-				}
-				resUsed := allocated[nodeID][resName]
+		if !ok {
```
Tal-or, line 165: There is also the case, which is not cover here, where you have a ```resCounters``` object i.e. ```ok == true```, but yet you don't have any allocatable CPUs (all reserved), thus the NRT object won't report the information correctly.
It's possible to have a  ```resCounters``` object without CPU's, for example you have an allocatable memory, but none CPUs.

fromanirh: Do we have real world example of such cases (or configuration possibly leading to these cases)?

fromanirh: Additionally, assuming we need to do this extra logic for cpus, don't we need to do the same logic at very least for memory? cc @swatisehgal

Tal-or: > Do we have real world example of such cases (or configuration possibly leading to these cases)?

It can happen on systems with very few CPUs (systems for testing, small VMs, etc.), for example, I encountered that on my testing system which has:
NUMA 0:
  CPU 0
  CPU 1
NUMA 1:
  CPU 2
  CPU 3
  
And I configured the first two CPUs as reserved.

fromanirh: ok. We need to somehow handle these corner cases but this looks like a configuration we should never seen in the (expected) production environments. Let's handle it in https://github.com/k8stopologyawareschedwg/resource-topology-exporter/pull/38

PR #36, opened 2021-09-15
report hugepages capacity per-numa-cell

fromanirh:
Introduce the sysinfo package to augment the ghw package, in order to report the per-numa-cell capacity.

Depends on https://github.com/k8stopologyawareschedwg/resource-topology-exporter/pull/34


fromanirh:
WIP: unit tests are still missing

fromanirh:
unit tests added, not WIP anymore and ready for review

fromanirh:
rebased after https://github.com/k8stopologyawareschedwg/resource-topology-exporter/pull/34 merged

pkg/sysinfo/hugepages.go
```diff
@@ -0,0 +1,157 @@
...
+}
+
+func hpResourceNameFromSize(sizeKB int) string {
+	if sizeKB == HugepageSize2Mi {
+		return "hugepages-2Mi"
+	}
+	if sizeKB == HugepageSize1Gi {
+		return "hugepages-1Gi"
+	}
+	// fallback, should never happen
```
MarSik, line 155: Except in some non Intel platforms. PPC can support weird sizes.

pkg/sysinfo/hugepages.go
```diff
@@ -0,0 +1,157 @@
...
+
+func readIntFromFile(path string) (int, error) {
+	data, err := ioutil.ReadFile(path)
+	if err != nil {
+		return -1, err
+	}
+	return strconv.Atoi(strings.TrimSpace(string(data)))
+}
+
+func hpResourceNameFromSize(sizeKB int) string {
```
MarSik, line 148: I wonder if this is the best way, why not convert the symbolic name to the size instead? You could then just traverse the /sys dir and expose everything you find in there without caring about what sizes are supported.

pkg/sysinfo/hugepages.go
fromanirh: Do we support PPC atm? I'm not sure. I think we will need a review all across the stack if we want to add that support

pkg/sysinfo/hugepages.go
fromanirh: done

pkg/resourcemonitor/resourcemonitor.go
```diff
@@ -190,7 +196,9 @@ func (rm *resourceMonitor) Scan(excludeList ResourceExcludeList) (topologyv1alph
 				// TODO when https://github.com/jaypipes/ghw/pull/268 get merged we should account memory capacity as well
 			}
 			if strings.HasPrefix(resName.String(), v1.ResourceHugePagesPrefix) {
-				// TODO
+				if hpCapacity, ok := hugepagesCapacity(hpCounters, resName.String(), nodeID); ok {
```
fromanirh, line 199: do we need to do this at each cycle?

swatisehgal (commented):
Looks overall good to me but I think we might have to refresh allocatable in case of hugepages as well

pkg/resourcemonitor/resourcemonitor.go
```diff
@@ -131,6 +132,11 @@ func (rm *resourceMonitor) Scan(excludeList ResourceExcludeList) (topologyv1alph
 		}
 	}
 
+	hpCounters, err := sysinfo.GetHugepageCounters(sysinfo.Handle{}, sysinfo.GetHugepages)
```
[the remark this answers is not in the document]
fromanirh: because it is assumed static the same way is handled by the kubelet. But we can be a bit smarter indeed.

pkg/resourcemonitor/resourcemonitor.go
```diff
@@ -46,10 +46,10 @@ type ResourceExcludeList struct {
...
 
 type Args struct {
-	Namespace          string
-	SysfsRoot          string
-	ExcludeList        ResourceExcludeList
-	RefreshAllocatable bool
+	Namespace            string
+	SysfsRoot            string
+	ExcludeList          ResourceExcludeList
+	RefreshNodeResources bool
```
fromanirh, line 53: it's very debatable that this change breaks the commandline tests (`main_test.go`), but let's address on a different PR.

swatisehgal:
We need to double check if hugepage capacity can be exposed to kubelet on the fly or it needs to be restarted before any updates in the configured number or hugepages are reflected in the node capacity. Lets add a comment corresponding to that in a future PR. Thanks for all the work in helping to improve RTE significantly!

/lgtm

fromanirh:
> We need to double check if hugepage capacity can be exposed to kubelet on the fly or it needs to be restarted before any updates in the configured number or hugepages are reflected in the node capacity. Lets add a comment corresponding to that in a future PR. Thanks for all the work in helping to improve RTE significantly!
> 
> /lgtm
absolutely. We need to document somewhere, and RTE docs is a nice place, how kubelet works. In this case I think we are either up to speed or even more advanced than the kubelet, so we are good to go indeed.

PR #38, opened 2021-09-20
WIP: account for all resources reserved on a NUMA zone

fromanirh:


fromanirh:
known broken, that's why is WIP :)

fromanirh:
I'm not convinced this is a reproducible issue. We will resume if/when we get a reproducer.

PR #42, opened 2021-10-05
gh actions: push to quay

fromanirh:
Add code to push images to quay


fromanirh:
still need to find a way to test before to merge, so let's go with it

PR #48, opened 2021-10-14
E2E test tuning for RTE d/s repo

Tal-or:
We want to reuse the same E2E for the RTE repo under openshift-kni org.
For that we need to do small adaptations in order to be able to vendor and compile the E2E test under the openshift-kni org

In this commit we pulled out all constants from the e2e_test.go file since it's not being vendor, thus the code does not compile since it needs those constants.


fromanirh (commented):
/lgtm
This is a good generalization which makes the testsuite easier to consume. Having a testsuite easier to consume means that is easier for projects that consume `rte` pkgs to have better test coverage, which is good, regardless of the specific project which wants to consume this code.

PR #49, opened 2021-10-21
e2e: modularize the suite

ffromani:
In this PR we modularize the test suite to make it as easy
as possible to consume.
This will allow the external project to consume the suite.
More projects consuming our suite will make it run more often,
so we will have more coverage, so we will have less bugs.

This PR is only about renaming and code moving, so the
chances of breakage are actually pretty low.


fromanirh:
before this PR (https://github.com/k8stopologyawareschedwg/resource-topology-exporter/actions/runs/1311273472):
image
after this PR (https://github.com/k8stopologyawareschedwg/resource-topology-exporter/actions/runs/1368746285):
image

PR #58, opened 2021-10-31
Assign test pods using node selectors

Tal-or:
During the e2e we deploy guaranteed pods that request some amount of resources and expecting to see the changes under the NRT object.
The problem is that on a cluster with multiple worker nodes, we don't know on which node the pod will land, and therefore we don't know which NRT object will change.

In order to solve the problem, we'll label an arbitrary worker node, with a unique label, at the beginning of the tests and use that label as a NodeSelector for all the pods.

### DIFF
--- a/test/e2e/utils/nodes/nodes.go
+++ b/test/e2e/utils/nodes/nodes.go
@@ -19,6 +19,8 @@ package nodes
 import (
 	"context"
 	"fmt"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/json"
 
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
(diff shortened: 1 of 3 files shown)


fromanirh (approved):
This is a good change in general, so let's have it anyway

test/e2e/utils/nodes/nodes.go
```diff
@@ -40,8 +42,16 @@ const (
 	LabelRole = "node-role.kubernetes.io"
 	// LabelHostname contains the key for the hostname label
 	LabelHostname = "kubernetes.io/hostname"
+
+	TestNodeLabel = "rte-e2e-test-node"
```
fromanirh, line 46: this constant is used here and in `pods.go`. The two modules should not import each other (if this is already the case, we should fix it in the future). Let's thus move this contant, and possibly others, in another package in a future PR.

test/e2e/utils/nodes/nodes.go
```diff
@@ -88,3 +98,29 @@ func FilterNodesWithEnoughCores(nodes []v1.Node, cpuAmount string) ([]v1.Node, e
...
+		Op:    "replace",
+		Path:  "/metadata/labels",
+		Value: labelsMap,
+	}}
+	payloadBytes, err := json.Marshal(patchPayload)
+	if err != nil {
+		return err
+	}
+
+	_, err = f.ClientSet.CoreV1().Nodes().Patch(context.TODO(), node.Name, types.JSONPatchType, payloadBytes, metav1.PatchOptions{})
```
fromanirh, line 121: just `return err` here, saving 4 lines :)

fromanirh (commented):
/lgtm

PR #59, opened 2021-10-31
sysinfo: fix hugepages information

cynepco3hahue:
The pod resource API returns hugepages amount in bytes, so
we should calculate the total amount of hugepages also in bytes.

Signed-off-by: Artyom Lukianov <alukiano@redhat.com>

### DIFF
--- a/pkg/sysinfo/sysinfo.go
+++ b/pkg/sysinfo/sysinfo.go
@@ -60,7 +60,7 @@ func GetMemoryResourceCounters(hnd Handle) (map[string]PerNUMACounters, error) {
 			numaDevs = make(PerNUMACounters)
 		}
 
-		numaDevs[hpage.NodeID] = int64(hpage.Total)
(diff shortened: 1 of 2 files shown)


pkg/sysinfo/sysinfo.go
```diff
@@ -60,7 +60,7 @@ func GetMemoryResourceCounters(hnd Handle) (map[string]PerNUMACounters, error) {
 			numaDevs = make(PerNUMACounters)
 		}
 
-		numaDevs[hpage.NodeID] = int64(hpage.Total)
+		numaDevs[hpage.NodeID] = int64(hpage.Total * hpage.SizeKB * 1024)
```
fromanirh, line 63: could you please elaborate in the commit message why this is needed?

cynepco3hahue: done

PR #60, opened 2021-11-01
Fix E2E test for metrics

Tal-or:
1. Get the correct Namespace for the test
2. Fix selector creation
With the previous implementation we ended up with the error shown here:
openshift-kni/resource-topology-exporter#28 (comment)
This patch fix that


fromanirh (commented):
/lgtm
/approve

PR #61, opened 2021-11-01
Move common constants into a separate package

Tal-or:

### DIFF
--- a/test/e2e/topology_updater/topology_updater.go
+++ b/test/e2e/topology_updater/topology_updater.go
@@ -26,8 +26,6 @@ import (
 	"fmt"
 	"time"
 
-	"github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/topology/v1alpha1"
-	topologyclientset "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/generated/clientset/versioned"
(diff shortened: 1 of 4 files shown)


test/e2e/utils/consts.go
```diff
@@ -0,0 +1,14 @@
+package utils
```
fromanirh, line 1: I like the idea but `utils` is an anti-pattern. Let's think of a better alternative.

Tal-or: How about ```testconsts```?

Tal-or: @fromanirh ^

fromanirh: on second thought the arguments with prevail is avoid kitchen-sink packages.
`testconsts` is a tiny improvement, but moving all the constants into a separate package is a sidegrade after all.
I think the best course of action is to repurpose this PR to add the 
```
// TestNodeLabel unique label for the node that
	TestNodeLabel = "rte-e2e-test-node"
```
constant, and leave the rest of the constants in the `nodes` package. All the constants belong to the node area after all, and overall keeping them there is the lesser evil and a cleanish solution.

Tal-or: Done

fromanirh (approved):

PR #62, opened 2021-11-03
Fix additional E2E tests errors

Tal-or:
- Fix nil dereference pointer errors
- Do not check for a single RTE pod in the cluster


fromanirh (commented):
/lgtm
/approve

PR #64, opened 2021-11-03
Remove obsolete tests from gh actions

Tal-or:


fromanirh:
these tests are not obsolete: the other way around, we planned initially a set of test which did not materialize to date, and it's still unclear they ever will. Any way, it's good to remove them to clean up a bit.

PR #65, opened 2021-11-04
WIP: Podreadiness: Inject custom pod status condition

Tal-or:
The update will allow us to report errors as Kubernetes objects using the Podreadiness feature:
https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-readiness-gate

We now have the following new conditions:
1. ResourcesScanned - means that resources scanned successfully.
2. TopologyUpdated - means that noderesourcetopology objects updated successfully.

We might have more conditions in the future, it is open for discussion.

WIP: e2e-test are missing


pkg/podreadiness/conditioninjector.go
```diff
@@ -0,0 +1,67 @@
...
+)
+
+type ConditionInjector struct {
+	cs      *kubernetes.Clientset
+	ns      string
+	podName string
+}
+
+func NewConditionInjector(cs *kubernetes.Clientset) (*ConditionInjector, error) {
+	nsVal, ok := os.LookupEnv("REFERENCE_NAMESPACE")
```
fromanirh, line 21: the naming is a bit unusual, I often saw `MY_POD_NAMESPACE` (and `MY_POD_NAME`) probably because almost everyone lazily mirrors the docs/code examples.

pkg/resourcetopologyexporter/resourcetopologyexporter.go
```diff
@@ -44,15 +46,25 @@ func Execute(cli podresourcesapi.PodResourcesListerClient, nrtupdaterArgs nrtupd
 		return err
 	}
 
-	resMon, err := NewResourceMonitor(cli, resourcemonitorArgs, rteArgs)
+	cs, err := nrtupdater.GetK8sClient("")
+	if err != nil {
+		return err
+	}
+
+	condIn, err := podreadiness.NewConditionInjector(cs)
```
fromanirh, line 54: please add a command line flag to turn out this behaviour - which should otherwise be on by default

pkg/resourcetopologyexporter/resourcetopologyexporter.go
```diff
@@ -173,8 +200,10 @@ func (rm *ResourceMonitor) Run(eventsChan <-chan PollTrigger) (<-chan nrtupdater
 
 				if err != nil {
 					klog.Warningf("failed to scan pod resources: %w\n", err)
+					rm.UpdateCondition(v1.ConditionFalse)
 					continue
 				}
+				rm.UpdateCondition(v1.ConditionTrue)
```
fromanirh, line 206: I'm not sure we should do this synchronously. I'd be happier if we update the condition asynchronously in a goroutine

fromanirh (commented):
the basic infrastructure and the overall layout seems solid. I'm not sure yet we want to do synchronous upgrades, which triggers some preemptive bells. Other than that, we will probably need some minor tuning to the names and to the conditions, but we'are in a pretty good shape already.

pkg/podreadiness/conditioninjector.go
Tal-or: I didn't write it. it was already defined under the DaemonSet definition.

pkg/resourcetopologyexporter/resourcetopologyexporter.go
Tal-or: Yes. I remember. In progress

pkg/resourcetopologyexporter/resourcetopologyexporter.go
Tal-or: Sure

pkg/podreadiness/conditioninjector.go
fromanirh: Right. Then, it's good as it is now.

pkg/nrtupdater/nrtupdater.go
```diff
@@ -108,16 +112,33 @@ func (te *NRTUpdater) Update(info MonitorInfo) error {
...
 
+func (te *NRTUpdater) UpdateCondition(status v1.ConditionStatus) {
+	cond := podreadiness.NewCondition(podreadiness.TopologyUpdated, status)
+
+	if status == v1.ConditionFalse {
+		cond.Reason = "UpdateFailed"
+		cond.Message = "failed to update noderesourcetopology object"
+	}
+
+	if err := te.condInj.Inject(cond); err != nil {
```
Tal-or, line 123: @fromanirh The ```Inject``` function being called by two different goroutine:
1. ```NRTUpdater.Run``` -> ```UpdateCondition``` -> ```Inject```
2. ```ResourceMonitor.Run``` -> ```UpdateCondition``` -> ```Inject```

Do I need to place Mutex inside the ```Inject``` function in order to avoid race conditions?

fromanirh: This is another great reason to move the podreainess handling code in its own goroutine :) This way we don't need explicit syncs, the channel takes care.

Tal-or: Yes. I'm working on that, I had something in mind, but ok I'll rearrange the code a bit

pkg/nrtupdater/nrtupdater.go
```diff
@@ -108,15 +110,28 @@ func (te *NRTUpdater) Update(info MonitorInfo) error {
 	return nil
 }
 
-func (te *NRTUpdater) Run(infoChannel <-chan MonitorInfo) chan<- struct{} {
+func (te *NRTUpdater) NewCondition(status v1.ConditionStatus) v1.PodCondition {
```
fromanirh, line 113: it seems you are never using `te` in the body of the method. So let's make this method a free function.

pkg/nrtupdater/nrtupdater.go
```diff
@@ -125,6 +140,9 @@ func (te *NRTUpdater) Run(infoChannel <-chan MonitorInfo) chan<- struct{} {
 				if te.args.Oneshot {
 					break
 				}
+				if condChan != nil {
```
fromanirh, line 143: this is good. A (IMHO) better approach is to provide a helper function like

```
package podreadines

...

func Set(cond v1.PodCondition) {
   if condChan != nil {
     condChan <- cond
  }
}
```
This way:
1. we factor the mechanism to set the condition into the relevant package (the fact we use a goroutine is a implementation detail
2. client code is not concerned about how the setting happens, just signals the intention to set.

pkg/podreadiness/conditioninjector.go
```diff
@@ -0,0 +1,79 @@
...
+		pod.Status.Conditions = conds
+	}
+
+	klog.Infof("pod conditions: %v", pod.Status.Conditions)
+
+	_, err = ci.cs.CoreV1().Pods(ci.ns).UpdateStatus(context.TODO(), pod, metav1.UpdateOptions{})
+	return err
+}
+
+func (ci *ConditionInjector) Run(condChan <-chan v1.PodCondition) {
```
fromanirh, line 69: we usually prefer to move the goroutine into the calling site:
```
func Run() {
  for {
      // infinite loop
  }
}

...

go podreadiness.Run()
```

pkg/resourcetopologyexporter/resourcetopologyexporter.go
```diff
@@ -44,19 +47,34 @@ func Execute(cli podresourcesapi.PodResourcesListerClient, nrtupdaterArgs nrtupd
...
-	resMon, err := NewResourceMonitor(cli, resourcemonitorArgs, rteArgs)
+	cs, err := nrtupdater.GetK8sClient("")
+	if err != nil {
+		return err
+	}
+
+	var condChan chan v1.PodCondition
+	if rteArgs.PodReadinessEnable {
+		condChan = make(chan v1.PodCondition)
+		condIn, err := podreadiness.NewConditionInjector(cs)
```
fromanirh, line 58: the `ConditionInjector` is the sole owner of the `clientset` so it should probably create itself

manifests/resource-topology-exporter-rbac.yaml
```diff
@@ -12,6 +12,12 @@ rules:
 - apiGroups: ["topology.node.k8s.io"]
   resources: ["noderesourcetopologies"]
   verbs: ["create", "update", "get", "list"]
+- apiGroups: [""]
+  resources: ["pods"]
+  verbs: ["get"]
+- apiGroups: [""]
+  resources: ["pods/status"]
+  verbs: ["update"]
```
fromanirh, line 20: very good! nice work

fromanirh (commented):
Initial review. Very good start, didn't look deeply in the end to end tests but at glance they look really nice.
We probably just need to reshuffle and repackage the application code.
Will do another pass later.

pkg/nrtupdater/nrtupdater.go
Tal-or: I see. But the function is not aware of the channel, shall I pass it as well?

pkg/podreadiness/conditioninjector.go
Tal-or: Interesting, because I saw the same technique here: https://github.com/k8stopologyawareschedwg/resource-topology-exporter/blob/master/pkg/nrtupdater/nrtupdater.go#L111

pkg/resourcetopologyexporter/resourcetopologyexporter.go
Tal-or: Ack

pkg/nrtupdater/nrtupdater.go
Tal-or: @fromanirh ^

pkg/podreadiness/conditioninjector.go
Tal-or: @fromanirh ^

pkg/nrtupdater/k8shelpers.go
```diff
@@ -27,3 +28,24 @@ func GetTopologyClient(kubeConfig string) (*topologyclientset.Clientset, error)
...
+	if kubeConfig == "" {
+		config, err = restclient.InClusterConfig()
+	} else {
+		config, err = clientcmd.BuildConfigFromFlags("", kubeConfig)
+	}
+	if err != nil {
+		return nil, err
+	}
+
+	cs, err := kubernetes.NewForConfig(config)
```
fromanirh, line 46: or just
`return kubernetes.NewForConfig(config)`

pkg/nrtupdater/nrtupdater.go
fromanirh: in general, better to hide the channel. So let's pass around a new type:
```
type Injector struct {}

...

func (inj *Injector) Set(cond v1.PodCondition) {}

...
```

pkg/podreadiness/conditioninjector.go
fromanirh: there is no universal recipe. Let's keep your code as-is and let's have a sweep later to make the usage pattern consistent.

fromanirh:
@Tal-or please rebase, and let's fix the clientset ownership and let's make NewCondition a free function. I'll re-evaluate the overall design and we will move forward with the goroutine placement.

Tal-or:
> @Tal-or please rebase, and let's fix the clientset ownership and let's make NewCondition a free function. I'll re-evaluate the overall design and we will move forward with the goroutine placement.

Done

pkg/podreadiness/condition.go
```diff
@@ -0,0 +1,27 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// RTEConditionType is a valid value for PodCondition.Type
+type RTEConditionType string
+
+// These are valid conditions of RTE pod.
+const (
+	// ResourcesScanned means that resources scanned successfully.
+	ResourcesScanned RTEConditionType = "ResourcesScanned"
```
fromanirh, line 16: let's rename this `PodresourcesFetched`

pkg/podreadiness/condition.go
```diff
@@ -0,0 +1,27 @@
...
+	// TopologyUpdated means that noderesourcetopology objects updated successfully.
+	TopologyUpdated RTEConditionType = "TopologyUpdated"
```
fromanirh, line 18: and let's rename this `NodeTopologyUpdated`

pkg/resourcetopologyexporter/resourcetopologyexporter.go
```diff
@@ -156,9 +170,21 @@ func NewResourceMonitor(cli podresourcesapi.PodResourcesListerClient, args resou
 	}, nil
 }
 
-func (rm *ResourceMonitor) Run(eventsChan <-chan PollTrigger) (<-chan nrtupdater.MonitorInfo, chan<- struct{}) {
+func NewCondition(status v1.ConditionStatus) v1.PodCondition {
```
fromanirh, line 173: we have the condition types already explicitely listed in the `podreadiness` package, so let's move this (and its sibling function in `nrtupdater`) there as well. Once we did this move, we would probably inline `NewConditionTemplate` in the calling sites. Like:
```
package podreadiness

import (
	"time"

	v1 "k8s.io/api/core/v1"
)

// RTEConditionType is a valid value for PodCondition.Type
type RTEConditionType string

// These are valid conditions of RTE pod.
const (
	// ResourcesScanned means that resources scanned successfully.
	ResourcesScanned RTEConditionType = "PodresourcesFetched"
	// TopologyUpdated means that noderesourcetopology objects updated successfully.
	TopologyUpdated RTEConditionType = "NodeTopologyUpdated"
)

func NewNodeTopologyUpdatedCondition(status v1.ConditionStatus) v1.PodCondition {
	cond := newConditionTemplate(TopologyUpdated, status)
	if status == v1.ConditionFalse {
		cond.Reason = "UpdateFailed"
		cond.Message = "failed to update noderesourcetopology object"
	}
	return cond
}

func NewPodresourcesFetchedCondition(status v1.ConditionStatus) v1.PodCondition {
	cond := newConditionTemplate(ResourcesScanned, status)

	if status == v1.ConditionFalse {
		cond.Reason = "ScanFailed"
		cond.Message = "failed to scan pod resources"
	}

	return cond
}

func newConditionTemplate(condType RTEConditionType, status v1.ConditionStatus) (condition v1.PodCondition) {
	return v1.PodCondition{
		Type:               v1.PodConditionType(condType),
		Status:             status,
		LastTransitionTime: metav1.Time{Time: time.Now()},
	}
}
```

pkg/nrtupdater/nrtupdater.go
fromanirh: OK, what about this?
we change `podreadiness` like this
```
func SetCondition(condChan chan<- v1.PodCondition, condType RTEConditionType, condStatus v1.ConditionStatus) {
	if condChan == nil {
		return
	}
	cond := newConditionTemplate(condType, condStatus)
	if condStatus == v1.ConditionFalse {
		switch condType {
		case ResourcesScanned:
			cond.Reason = "ScanFailed"
			cond.Message = "failed to scan pod resources"
		case TopologyUpdated:
			cond.Reason = "UpdateFailed"
			cond.Message = "failed to update noderesourcetopology object"
		}
	}
	condChan <- cond
}

func newConditionTemplate(condType RTEConditionType, status v1.ConditionStatus) (condition v1.PodCondition) {
	return v1.PodCondition{
		Type:               v1.PodConditionType(condType),
		Status:             status,
		LastTransitionTime: metav1.Time{Time: time.Now()},
	}
}
```

and we use like this:
```
podreadiness.SetCondition(condChan, podreadiness.TopologyUpdated, status)
...
podreadiness.SetCondition(condChan, podreadiness.TopologyUpdated, condStatus)
```

This solves my concern and it's not a big shakeup of the current architecture

pkg/resourcetopologyexporter/resourcetopologyexporter.go
Tal-or: We don't need the ```NewPodresourcesFetchedCondition``` and ```NewNodeTopologyUpdatedCondition``` if we're using your previous suggestion:
https://github.com/k8stopologyawareschedwg/resource-topology-exporter/pull/65#discussion_r751100455
Because we're updating the ```Reason``` and ```Message``` in the switch case.
But the general suggestion of moving everything into the ```podreadiness``` sounds fair to me.

fromanirh: this comment is indeed obsoleted by https://github.com/k8stopologyawareschedwg/resource-topology-exporter/pull/65#discussion_r751100455

Tal-or: **Deleted the comment by mistake:** We don't need the NewPodresourcesFetchedCondition and NewNodeTopologyUpdatedCondition if we're using your previous suggestion:
#65 (comment)
Because we're updating the Reason and Message in the switch case.
But the general suggestion of moving everything into the podreadiness sounds fair to me.

pkg/podreadiness/condition_test.go
```diff
@@ -0,0 +1,58 @@
...
+			name:     "PodresourcesFetched condition is false",
+			status:   v1.ConditionFalse,
+			cType:    PodresourcesFetched,
+			expected: newConditionTemplate(NodeTopologyUpdated, v1.ConditionTrue),
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			go SetCondition(c, tc.cType, tc.status)
```
Tal-or, line 48: Is running a go-routine in unit-tests is acceptable?

Tal-or: My only concern is from blocking the unit test but in such an extreme case I think this should be our least concern.

PR #66, opened 2021-11-11
rte: report the proper tmPolicy

fromanirh:
For single-numa-node (and in general) we should not report the tm policy
straight away, but we need to take into account the tm scope as well.
Otherwise, the scheduler plugin cannot cope with the data we provide.


cmd/resource-topology-exporter/main.go
```diff
@@ -168,6 +169,14 @@ func defaultTopologyManagerPolicy() string {
 	return ""
 }
 
+func defaultTopologyManagerScope() string {
+	if val, ok := os.LookupEnv("TOPOLOGY_MANAGER_SCOPEt"); ok {
```
fromanirh, line 173: typo

PR #67, opened 2021-11-11
gh actions: release: fix the release flow

fromanirh:

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -18,6 +18,7 @@ jobs:
     runs-on: ubuntu-20.04
     env:
       RTE_CONTAINER_IMAGE: quay.io/k8stopologyawareschedwg/resource-topology-exporter
+      E2E_TOPOLOGY_MANAGER_POLICY: SingleNUMANodeContainerLevel
     steps:
     - name: checkout sources
       uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -18,6 +18,7 @@ jobs:
     runs-on: ubuntu-20.04
     env:
       RTE_CONTAINER_IMAGE: quay.io/k8stopologyawareschedwg/resource-topology-exporter
+      E2E_TOPOLOGY_MANAGER_POLICY: SingleNUMANodeContainerLevel
```
Tal-or, line 21: Something is very suspicious, how the test passed if it is looking for:
https://github.com/k8stopologyawareschedwg/resource-topology-exporter/blob/88595fb0d865dedcf5a0333b977f1731d2ed0fa7/test/e2e/utils/nodetopology/nodetopology.go#L53

fromanirh: Good point, let's look into it a bit deeper

fromanirh: Previously, we were reading the expected value from the kubelet config file, so we required no env variable. This obviously works only on kind, we fixed this between 0.2.5 and now.

fromanirh:
@Tal-or  any other comment on this PR?

Tal-or:
> @Tal-or any other comment on this PR?

Nope. Just took a second look at the implementation.
/lgtm

PR #69, opened 2021-11-12
topologypolicy: resync pkg with NFD

fromanirh:
We can get rid of in-house types and use the ones from the recent CRD
version, like NFD is already doing.
And add unit tests along the way.


swatisehgal:
Thanks for making this change to RTE. Makes sense to ensure that both NFD and RTE determine the topologymanager policy and subsequently its population in the CR in the same way.

/lgtm

Issue #71, opened 2021-11-12
[test] deflake resourcemonitor test


fromanirh:
a resurcemonitor test is flaky:

https://github.com/k8stopologyawareschedwg/resource-topology-exporter/actions/runs/1452430869/attempts/1

We should fix it, making it fully reliable.

fromanirh:
solved by https://github.com/k8stopologyawareschedwg/resource-topology-exporter/pull/72

PR #72, opened 2021-11-15
deflake resourcemonitor test

jlojosnegros:
#71 
TestNormalizeContainerDevices was  flaky due to a typo in the sort functions.


fromanirh:
when comparing device identifiers pertaining to a given resource, the test code was incorrectly iterating twice over the same resource instead of iterating once on each list of the pair being checked, right?

If so, please add this sentence to the commit message, to avoid our future selves to remember quickly where the typo was.

fromanirh (approved):
thanks!

PR #79, opened 2021-11-22
e2e: add `_test` suffix to test files

fromanirh:
Trivial rename only
Emphasize that these files contains test code.


fromanirh:
/hold
I need to review how the suffix, TestMain and `go test -c` all play together

fromanirh:
we will re-evaluate later

PR #80, opened 2021-11-22
e2e: remove hardcoded timeouts

fromanirh:
We should wait relatively to the poll interval.
Removing hardcoded settings make the test more robust and
easier to adapt, and easier to consume for external projects.

### DIFF
--- a/test/e2e/rte/rte.go
+++ b/test/e2e/rte/rte.go
@@ -41,11 +41,13 @@ import (
 	e2enodetopology "github.com/k8stopologyawareschedwg/resource-topology-exporter/test/e2e/utils/nodetopology"
 	e2epods "github.com/k8stopologyawareschedwg/resource-topology-exporter/test/e2e/utils/pods"
 	e2etestconsts "github.com/k8stopologyawareschedwg/resource-topology-exporter/test/e2e/utils/testconsts"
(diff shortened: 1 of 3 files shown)


test/e2e/rte/rte.go
```diff
@@ -130,7 +139,7 @@ var _ = ginkgo.Describe("[RTE][InfraConsuming] Resource topology exporter", func
 					return false
 				}
 				return reason == nrtupdater.RTEUpdateReactive
-			}, 5*time.Second, 1*time.Second).Should(gomega.BeTrue(), "didn't get updated node topology info")
```
Tal-or, line 133: This ```timeout``` value is not a multiple of ```poll interval``` value (the default - and in our case we're using the default is ```10s```) so I'm not sure I understand why the new implementation is better.
What am I missing here?

Tal-or:
Most of the tests were skipped so we can't conclude we improve the tests. Did you notice better results under your local setup?

test/e2e/rte/rte.go
fromanirh: I missed to update this part :)

fromanirh:
> Most of the tests were skipped so we can't conclude we improve the tests. Did you notice better results under your local setup?

We cannot indeed. This is about flakiness prevention. Runned locally and/or on multi-worker cluster.

PR #81, opened 2021-11-24
e2e: version: switch to semver pkg

fromanirh:
We are more explicit about the version we want to parse,
and the license is CNCF-friendlier.
No change in behaviour, we use this pkg only in the tests anyway.


fromanirh:
obsolete

PR #82, opened 2021-11-24
test: e2e: avoid hardcoded value, introspect pod

fromanirh:
We use hardcoded value in the tests mostly (= 95% of the cases)
just because it's the simplest way. A more robust and flexible
way is to introspect the actual pod and extract the values
we care about. After all, there often is a set of correct values,
and we care more about state matching expectations (with again
a set of correct values, not a single one) rather than a exact check.

### DIFF
--- a/test/e2e/utils/pods/rtepod/rtepod.go
+++ b/test/e2e/utils/pods/rtepod/rtepod.go
@@ -1,0 +1,102 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// rtepod includes utilities to introspect the RTE pod - and the RTE pod only
+package rtepod
+
+import (
+	"fmt"
+	"strconv"
+	"strings"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+const (
+	rteExecutable = "resource-topology-exporter"
+
+	notificationFileOption = "--notify-file"
+)
+
+func FindMetricsPort(rtePod *corev1.Pod) (int, error) {
+	for idx := 0; idx < len(rtePod.Spec.Containers); idx++ {
+		cnt := rtePod.Spec.Containers[idx] // shortcut
+		if !isRTEContainer(cnt) {
+			continue
+		}
+
+		for _, envVar := range cnt.Env {
+			if envVar.Name == "METRICS_PORT" {
+				val, err := strconv.Atoi(envVar.Value)
+				if err != nil {
+					return 0, err
+				}
+				return val, nil
+			}
+		}
+	}
+	return 0, fmt.Errorf("cannot find METRICS_PORT environment variable")
+}
+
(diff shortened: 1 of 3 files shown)


test/e2e/utils/pods/rtepod/rtepod.go
```diff
@@ -0,0 +1,87 @@
...
+func FindNotificationFilePath(rtePod *corev1.Pod) (string, error) {
+	for idx := 0; idx < len(rtePod.Spec.Containers); idx++ {
+		cnt := &rtePod.Spec.Containers[idx] // shortcut
+		for _, arg := range cnt.Command {
+			if strings.Contains(arg, notificationFileOption) {
+				return extractOptionValue(arg)
+			}
+		}
+		for _, arg := range cnt.Args {
+			if strings.Contains(arg, RTEExecutable) {
```
Tal-or, line 73: I think you meant ```notificationFileOption```?

test/e2e/utils/pods/rtepod/rtepod.go
```diff
@@ -0,0 +1,87 @@
...
+import (
+	"fmt"
+	"strconv"
+	"strings"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+const (
+	RTEExecutable = "resource-topology-exporter"
```
cynepco3hahue, line 29: probably for now we do not need to make it public

test/e2e/utils/pods/rtepod/rtepod.go
fromanirh: good point. Best would be check both

test/e2e/utils/pods/rtepod/rtepod.go
```diff
@@ -0,0 +1,87 @@
...
+
+	notificationFileOption = "--notify-file"
+)
+
+func FindMetricsPort(rtePod *corev1.Pod) (int, error) {
+	for _, envVar := range rtePod.Spec.Containers[0].Env {
```
cynepco3hahue, line 35: Can we always assume that the first container will be the RTE one? Maybe we can iterate over all containers and use one that have env var `METRICS_PORT`

test/e2e/utils/pods/rtepod/rtepod.go
```diff
@@ -0,0 +1,87 @@
...
+		}
+	}
+	return 0, fmt.Errorf("cannot find METRICS_PORT environment variable")
+}
+
+func FindRTEContainerName(rtePod *corev1.Pod) (string, error) {
+	for idx := 0; idx < len(rtePod.Spec.Containers); idx++ {
+		cnt := &rtePod.Spec.Containers[idx] // shortcut
+		for _, arg := range cnt.Command {
+			if strings.Contains(arg, RTEExecutable) {
```
cynepco3hahue, line 51: we still somehow related on hardcoded values, but I do not have a better solution :(

test/e2e/utils/pods/rtepod/rtepod.go
cynepco3hahue: nope, it first finds the relevant container and after extracts options from it.

cynepco3hahue (commented):
Some small comments, otherwise lgtm

test/e2e/utils/pods/rtepod/rtepod.go
fromanirh: indeed, I overlooked that part. Will fix.

test/e2e/utils/pods/rtepod/rtepod.go
fromanirh: this is true. Howver I believe this PR is still a worthy improvement because:

1. we depend on less values, and more and more stable: we don't change the executable name often (ort at all) and same is true for the name of the options
2. if we ever do, now the tests should break with a clear indication why, so we can fix/revisit

PR #85, opened 2021-12-01
Throttle RTE events

jlojosnegros:
Throttle events to avoid storms of podresources pools and
NRT objects updates because both are not cheap operations.


cynepco3hahue:
@jlojosnegros PR looks good to me, did you find a way to test it?

pkg/ratelimit/ratelimit.go
```diff
@@ -0,0 +1,82 @@
+package ratelimit
```
fromanirh, line 1: please add the license blurb like other source files in this repo

pkg/ratelimit/ratelimit.go
```diff
@@ -0,0 +1,82 @@
...
+
+func NewWithEPS(eventsPerSecond int64, src <-chan Event) *RateLimiter {
+	rt := newCommon(src)
+	rt.rt = ratelimit.New(int(eventsPerSecond))
+	return rt
+}
+
+func NewUnlimited(src <-chan Event) *RateLimiter {
+	rt := newCommon(src)
+	rt.rt = ratelimit.NewUnlimited()
```
fromanirh, line 31: considering the uber `ratelimit` package seems to provide all the functionalities we need, can we perhaps simplify our wrapping code?

pkg/ratelimit/ratelimit.go
```diff
@@ -0,0 +1,82 @@
...
+		ch:           make(chan Event),
+		outputCh:     make(chan Event),
+		doneSender:   make(chan struct{}),
+		doneReceiver: make(chan struct{}),
+	}
+}
+
+func (rt *RateLimiter) Run() {
+
+	sender := func() {
```
fromanirh, line 47: I'm a bit surprised we need two helper goroutines.
Additionally, we seem to add some overhead in the unlimited usecase. Do we have any option to optimize the unlimited use case so we don't introduce overhead (or as low as possible?)

pkg/ratelimit/ratelimit.go
```diff
@@ -0,0 +1,82 @@
...
+	go sender()
+	go receiver()
+}
+
+func (rt *RateLimiter) Stop() {
+	rt.doneReceiver <- struct{}{}
+	rt.doneSender <- struct{}{}
+}
+
+func (rt RateLimiter) OuputChannel() <-chan Event {
```
fromanirh, line 80: not sure we need a wrapper method: exposing channels as part of the API is a pretty standard practice in golang.

cmd/resource-topology-exporter/main.go
```diff
@@ -112,6 +112,7 @@ func parseArgs(args ...string) (ProgArgs, error) {
 	refCnt := flags.String("reference-container", "", "Reference container, used to learn about the shared cpu pool\n See: https://github.com/kubernetes/kubernetes/issues/102190\n format of spec is namespace/podname/containername.\n Alternatively, you can use the env vars REFERENCE_NAMESPACE, REFERENCE_POD_NAME, REFERENCE_CONTAINER_NAME.")
 
 	flags.StringVar(&pArgs.RTE.NotifyFilePath, "notify-file", "", "Notification file path.")
+	flags.Int64Var(&pArgs.RTE.MaxEventsPerSecond, "max-events-per-second", 1, "Max times per second resources will be scanned and updated")
```
fromanirh, line 115: the default should be unlimited

fromanirh (commented):
nice start :) Indeed let's figure out a way to add some tests.
Unit (/integration) tests are fine. We don't _need_ e2e tests I believe, but they are of course welcome.

cmd/resource-topology-exporter/main.go
cynepco3hahue: hm, I asked to change it to 1, to avoid calling a lot of updates by default, IMHO one update per second is more than enough, in general, I preferred more granularity, like work with milliseconds but looks like it is not supported.

fromanirh: there is a tradeoff between frequency of updates, thus extra load on the node and most notably on the apiserver, and quality of the scheduling decisions. We need to explore this tradeoff.
I agree that most likely we will need some form of throttling, but this should be a per-cluster decision, thus enforced by either the deployer, the operator, or the cluster admin. Thus, I think the default should be unlimited.

fromanirh:
Additionally, since the first patch breaks the tests and the second fixes it, please squash them together

cmd/resource-topology-exporter/main.go
fromanirh: Additionally, it's likely we will need to have sub-second granularity (e.g. 100ms)

pkg/ratelimit/ratelimit.go
fromanirh: You can provide a channel allowing only one directional operation (e.g. only reads, only writes), and I think that's the right approach here https://gobyexample.com/channel-directions
example: https://pkg.go.dev/time#Ticker

fromanirh: bah, expressed myself poorly. I'll probably just provide an example to illustrate my point later.

pkg/ratelimit/ratelimit.go
jlojosnegros: I have two goroutines because I want to decouple input channel from output channel.
I want to give the client the oportunity to always be able to write into the input channel with almost no bloking time while writing in the output channel just following the marked time period, so that's why I have one routine to be always reading from "src" and trying to write into "ch" but not stopping if "ch" already have an item, and the sender one reads from "ch" and send to output...

pkg/ratelimit/ratelimit.go
jlojosnegros: do you mean to have a "private" read-write channel and a "public" read-only channel and assign the first one to the second one instead of using a method to do it?

jlojosnegros: type xxx struct {
    outputCh chan Event
    C <-chan Event
}

and eventually do:
C = outputCh

cmd/resource-topology-exporter/main.go
jlojosnegros: Regarding this ... do you prefer to have a fixed time base and let only the number of events as a configurable value o give the option to configure both of them?

pkg/ratelimit/ratelimit.go
jlojosnegros: ok

jlojosnegros:
I have changed the RateLimiter to use a blocker channel as input and I have therefore simplified the code a little bit.

I have been also testing the overload difference between the current solution and the one assigning "src" channel directly to "C" and there seems to use the same amount of time. I have left this one because the code is homogeneus between the limited and unlimited use case, but I can change it if you think it the other solution is better,.

cmd/resource-topology-exporter/main.go
```diff
@@ -112,7 +112,8 @@ func parseArgs(args ...string) (ProgArgs, error) {
 	refCnt := flags.String("reference-container", "", "Reference container, used to learn about the shared cpu pool\n See: https://github.com/kubernetes/kubernetes/issues/102190\n format of spec is namespace/podname/containername.\n Alternatively, you can use the env vars REFERENCE_NAMESPACE, REFERENCE_POD_NAME, REFERENCE_CONTAINER_NAME.")
 
 	flags.StringVar(&pArgs.RTE.NotifyFilePath, "notify-file", "", "Notification file path.")
-	flags.Int64Var(&pArgs.RTE.MaxEventsPerSecond, "max-events-per-second", 1, "Max times per second resources will be scanned and updated")
+	flags.Int64Var(&pArgs.RTE.MaxEventsPerTimeUnit, "max-events-per-time-unit", 1, "Max times per time unit that resources will be scanned and updated.\n See 'time-unit-to-limit-events' to set the base time unit")
+	flags.DurationVar(&pArgs.RTE.TimeUnitToLimitEvents, "time-unit-to-limit-events", 100*time.Millisecond, "Base time unit to calculate the maximum rate of events. 'max-events-per-time-unit' / 'time-unit-to-limit-events")
```
jlojosnegros, line 116: Not a big fan of the chosen name but I didn't found anything better.
Default value just as an example. 
Suggestions would be really appreciated for both name and default value.

pkg/ratelimit/ratelimit.go
```diff
@@ -0,0 +1,77 @@
...
+	rt := RateLimiter{
+		src:      src,
+		outputCh: make(chan Event),
+		doneCh:   make(chan struct{}),
+	}
+	rt.C = rt.outputCh
+	return &rt
+}
+
+func (rt *RateLimiter) Run() {
```
cynepco3hahue, line 60: Please correct me if I am wrong, but I do not see any blocking loops under the `Run` meaning, it will exit straight forward after the call and also will exit the go routine.

pkg/ratelimit/ratelimit.go
```diff
@@ -0,0 +1,77 @@
...
+)
+
+type Event struct {
+	Timestamp time.Time
+	Tag       int
+}
+
+type RateLimiter struct {
+	outputCh chan Event
+	C        <-chan Event
```
cynepco3hahue, line 31: What does `C` mean?

pkg/ratelimit/ratelimit_test.go
```diff
@@ -0,0 +1,119 @@
...
+		case <-readCh:
+			r := result{TsLastRcv: tsLastRcv, TsRcv: time.Now()}
+			tsLastRcv = time.Now()
+			results = append(results, r)
+		case <-sync:
+			finish = true
+		}
+	}
+
+	return results[1:]
```
cynepco3hahue, line 118: Why do you ignore the first result?

pkg/ratelimit/ratelimit_test.go
```diff
@@ -0,0 +1,119 @@
...
+func receiver(t *testing.T, readCh <-chan Event, sync <-chan struct{}) []result {
+	var results []result
+
+	tsLastRcv := time.Now()
+	finish := false
+	for !finish {
+		select {
+		case <-readCh:
+			r := result{TsLastRcv: tsLastRcv, TsRcv: time.Now()}
+			tsLastRcv = time.Now()
```
cynepco3hahue, line 111: where do you use this variable?

pkg/ratelimit/ratelimit.go
jlojosnegros: Yes, you are right, there is no blocking loops. I would have called this method "Start" instead of "Run" but I called "Run" to follow the names I have seen  in the RTE (ResourceObserver for example )

pkg/ratelimit/ratelimit.go
jlojosnegros: `C` is just a public read-only view of `outputCh` so `RateLimiter` client could read events from it.
I have used `C` as name following the trail of `Timer`. I though it was a common practice in go, but not sure.

pkg/ratelimit/ratelimit_test.go
```diff
@@ -0,0 +1,119 @@
...
+			results = append(results, r)
+		case <-sync:
+			finish = true
```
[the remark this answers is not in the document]
jlojosnegros: I usually do that but @fromanirh told me on a previous review that this is a better and smoother way to finish...

pkg/ratelimit/ratelimit_test.go
jlojosnegros: The test tries to ensure  always there is the same amount of time between output events to fulfill the rate condition.
The first event came out as soon as it is in the channel so the first period will be always "wrong" so to say. That's why I skip the first one.

pkg/ratelimit/ratelimit_test.go
jlojosnegros: tsLastRcv is used in the result struct to lately compute time difference between output events and be able to check ratio.

pkg/ratelimit/ratelimit.go
cynepco3hahue: I thought we should always run this loop and does not exit straightforward after the call.
IMHO we should change it to
```
func (rt *RateLimiter) Run() {
for {
			select {
			case event := <-rt.src:
				rt.rt.Take()
				rt.outputCh <- event
			case <-rt.doneCh:
				return
			}
		}
}
```
and under the `Execute` call it via
`go Run(...)`

pkg/ratelimit/ratelimit_test.go
cynepco3hahue: I see now, missed it

pkg/ratelimit/ratelimit_test.go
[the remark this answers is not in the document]
cynepco3hahue: @fromanirh, To be honest, does not really see a difference :) Just return or set some boolean and after return

pkg/ratelimit/ratelimit.go
jlojosnegros: Yep I have nothing against that change ... I just did it that way to keep some homogeneity with the current code.
If we think this is a better way to do it, just to left concurrency in the caller hands, maybe we should also change ResourceObserver to behave the same way.

fromanirh: yes, we can use some cleanup in this area for this project. Let's pause for a sec while we evaluate a cleanup PR before we move forward.

pkg/ratelimit/ratelimit_test.go
```diff
@@ -0,0 +1,119 @@
...
+limitations under the License.
+*/
+package ratelimit
+
+import (
+	"math"
+	"testing"
+	"time"
+
+	. "github.com/smartystreets/goconvey/convey"
```
fromanirh, line 23: not a fan of this style of tests, but once again good to be consistent. Let's cleanup later in one bang.

pkg/ratelimit/ratelimit_test.go
```diff
@@ -0,0 +1,119 @@
...
+
+		So(timeUnit, ShouldBeGreaterThanOrEqualTo, time.Microsecond)
+
+		sourceCh := make(chan Event)
+		sut := NewWithEPS(numberOfEvents, timeUnit, sourceCh)
+		sut.Run()
+
+		done := make(chan struct{})
+
+		Convey("When some messages are send", func() {
```
fromanirh, line 70: are `sent` ?

pkg/ratelimit/ratelimit_test.go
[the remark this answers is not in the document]
fromanirh: In retrospect it's true, there is not a big difference. I overestimated the magnitude of the improvement :) But at this point let's keep it like this, not worth changing it back again.

pkg/ratelimit/ratelimit_test.go
fromanirh: this information is worth to be captured in a comment above the `return` line

cmd/resource-topology-exporter/main.go
fromanirh: This is going to be hard to tune, so I'd just expose a `-max-events-per-second` flag and compute the values internally.
IOW, let's start with the simplest user-facing interface. We can keep the current internal interfaces.

pkg/ratelimit/ratelimit.go
fromanirh: I think this is fine. Maybe `C` is a bit too short, we can maybe rename it to `Events` for better clarity.

pkg/ratelimit/ratelimit_test.go
jlojosnegros: my fault

fromanirh:
thanks, seems good. Please squash your commits.

pkg/ratelimit/ratelimit.go
cynepco3hahue: My main concern is that we exit the Run straight forward after the call and close the blocking goroutine

cynepco3hahue: But I can see we already have https://github.com/k8stopologyawareschedwg/resource-topology-exporter/pull/94/files
so it ok for me.

fromanirh:
please rebase

jlojosnegros:
Structure has changed so much I do not think is worthy to rebase this code. 
I will adapt the solution to the new structure in a new PR

fromanirh:
> Structure has changed so much I do not think is worthy to rebase this code. I will adapt the solution to the new structure in a new PR

ok, works for me!

jlojosnegros:
Closing this pr due to changes in the baseline. See #95 for continuation.

pkg/ratelimit/ratelimit.go
fromanirh: circling back to this, decoupling input and output is fine, I'm just not sure we need two goroutines to achieve this; this may very well be the case, and so it needs a good description/rationale (code comments and/or commit message).
It may also be the case that the API we choose (exposing a channel) drives the implementation, e.g. making the two-goroutine approach the best; this is fine, provided that it is documented :)

PR #87, opened 2021-12-08
e2e: small fixes to tests

cynepco3hahue:
- condition tests: do not run ginkgo.Expect under Eventually because it will result
in drop us out of Eventually.
- topology updater tests: check the generation instead of resource version, because generation updated
only when was change under the non status field.

Signed-off-by: Artyom Lukianov <alukiano@redhat.com>

### DIFF
--- a/test/e2e/rte/conditions.go
+++ b/test/e2e/rte/conditions.go
@@ -69,40 +69,42 @@ var _ = ginkgo.Describe("[RTE][Monitoring] conditions", func() {
 		}
 	})
 
+	waitForPodCondition := func(podName string, conditionType podreadiness.RTEConditionType, expectedConditionStatus v1.ConditionStatus) bool {
+		pods, err := e2epods.GetPodsByLabel(f, namespace, fmt.Sprintf("name=%s", podName))
+		if err != nil {
+			return false
+		}
+
+		if len(pods) == 0 {
+			return false
+		}
+
+		return cmpConditionsByTypeAndStatus(pods[0].Status.Conditions, conditionType, expectedConditionStatus)
+	}
+
 	ginkgo.Context("with NRT objects created", func() {
 		ginkgo.It("should have custom RTE conditions under the pod status", func() {
-
 			gomega.Eventually(func() bool {
-				pods, err := e2epods.GetPodsByLabel(f, namespace, fmt.Sprintf("name=%s", e2etestenv.RTELabelName))
-				gomega.Expect(err).NotTo(gomega.HaveOccurred())
-				gomega.Expect(len(pods)).NotTo(gomega.BeZero())
-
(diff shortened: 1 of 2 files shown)


test/e2e/rte/conditions.go
```diff
@@ -69,40 +69,42 @@ var _ = ginkgo.Describe("[RTE][Monitoring] conditions", func() {
...
-				return cmpConditionsByTypeAndStatus(pods[0].Status.Conditions, podreadiness.NodeTopologyUpdated, v1.ConditionTrue)
+				return waitForPodCondition(e2etestenv.RTELabelName, podreadiness.NodeTopologyUpdated, v1.ConditionTrue)
 				// wait for twice the poll interval, so the conditions will have enough time to get updated
 			}, 2*timeout, 1*time.Second).Should(gomega.BeTrue(), "pod contains wrong condition value")
 		})
 
-		ginkgo.It("should change the RTE conditions under the pod status accordingly", func() {
+		// it is flaky test under the NRO, because it is possible that the operator will re-create the CRD, before the pod
+		// condition updated
+		ginkgo.It("[SKIP NRO] should change the RTE conditions under the pod status accordingly", func() {
```
fromanirh, line 100: please remove this change

test/e2e/topology_updater/topology_updater.go
```diff
@@ -199,7 +199,7 @@ var _ = ginkgo.Describe("[TopologyUpdater][InfraConsuming] Node topology updater
 					framework.Logf("failed to get the node topology resource: %v", err)
 					return false
 				}
-				return finalNodeTopo.ObjectMeta.ResourceVersion != initialNodeTopo.ObjectMeta.ResourceVersion
+				return finalNodeTopo.ObjectMeta.Generation != initialNodeTopo.ObjectMeta.Generation
```
fromanirh, line 202: nice catch

test/e2e/rte/conditions.go
cynepco3hahue: In general, I want to skip some tests that are flaky only under the NRO, and adding some tag is the easiest way to do it.

fromanirh: Adding labels is fine, but we should not tie to specific external projects. Could you use a more generic label please?

fromanirh: This specific test can probably use some work to make it less flaky, but this is totally stuff for another PR :)
Lacking better ideas, I'd suggest a label like `[EventChain]` to suggest that this test depends on a (complex) chain of events and thus is possibly flaky. Other tests may need this label as well (metrics and notification come to mind).

cynepco3hahue: sure will change it with the comment, in general it the single flaky that I saw under my environment all others was flaky because we started them before the RTE daemon set was ready(we started RTE tests before the install)

fromanirh (commented):
/lgtm
/approve

Issue #83, opened 2021-11-25
replace go-version


fromanirh:
In the e2e tests we use https://github.com/hashicorp/go-version which is a fine package but whose license is not auto-allowed by CNCF (see: https://github.com/cncf/foundation/blob/master/allowed-third-party-license-policy.md)
In order to be as much friendly as possible with CNCF, we should replace this dependency with something with a auto-allowed license, like https://github.com/aquasecurity/go-version

Finally, it may be needed to revisit how we auto-create the version (nowadays just using `git describe --tags`), which I'm not sure it's a well formed SEMVER anyway.

jlojosnegros:
# Regarding how we auto-create versions.
## Context
### Semver format
Semver versions use the format
```
MAJOR.MINOR.PATCH-pre_release+metadata
```
Here we are concern about `pre_release` and `metadata` fields mainly.

#### [pre-release](https://semver.org/#spec-item-9)
Identifiers MUST comprise only ASCII alphanumerics and hyphens [0-9A-Za-z-]

#### [metadata](https://semver.org/#spec-item-10)
Build metadata MUST be ignored when determining version precedence

### `git describe`
From [documentation](https://git-scm.com/docs/git-describe) `git describe --tags` command returns the closest reachable tag from a given commit and, if the tag does **not** point to the commit itself, it add some suffixes.
Ej: For a commit without a tag:
```bash
$> git describe --tags
v0.2.5-68-g91f2c330
```
Being:
- `v0.2.5`: Closest tag found
- `68`: Number of commits on top of the tag
- `g91f2c330`: hash of the last commit

## Problems

### Current tag format is not a valid semver
As we can see in [semver homepage](https://semver.org/#is-v123-a-semantic-version), using 'v' as prefix is not valid.
This seems a minor problem.

### Semver [precedence rules.](https://semver.org/#spec-item-11)
Here is the main problem.
With `git describe` command we build new versions adding a suffix to the tag, that is adding a `pre_release` field using semver terminology, so a version with a suffix is "greater" (newest) than the one without it.
```
v0.2.5 < v0.2.5-68-g91f2c330
```

However, as we can see in the semver precedence rules, any version with a `pre_release` field is considered "less" than (oldest) the one without it, because pre-released versions seems to be the working versions leading to the release of the version mentioned.

```
v0.2.5-68-g91f2c330 < v0.2.5 
```

## Possible solutions.
[...]

PR #89, opened 2021-12-09
Generate semver compatible versions from tags.

jlojosnegros:
Use git-semver tool to generate semver compatible versions for
development commits.

Add tools.go file to track dependency of git-semver as it is a tool just
used to build purposes.

See: issue #83


fromanirh:
```
Run RTE_CONTAINER_IMAGE=${RTE_CONTAINER_IMAGE} RUNTIME=docker make image
make: /home/runner/go/bin/git-semver-v1.0.0: Command not found
```

fromanirh:
I'm not sure we want to go down the `bin-go` path

jlojosnegros:
I think it is a nice way to keep the dependency tracked and easy to use without requiring any special operation neither locally nor in Ci.
If not using bin-go or something similar, should we go then for a shell based solution?

fromanirh:
> I think it is a nice way to keep the dependency tracked and easy to use without requiring any special operation neither locally nor in Ci. If not using bin-go or something similar, should we go then for a shell based solution?

considering how invasive (in this context) adding bin-go is, and considering we are unlikely to add extra deps (nothing else on the radar), let's evaluate how hard is to build `go-semver` locally and to use it directly

fromanirh:
like: https://github.com/k8stopologyawareschedwg/resource-topology-exporter/pull/90

jlojosnegros:
Well I do not think bin-go is so invasive cause It just helps automatizing the process to add or update a dep, but it is not needed for build them at all, but ok.  I will go with the #90 solution

PR #90, opened 2021-12-10
build: use git-semver to create the version

fromanirh:
git-semver can create a proper semantic version,
while `git describe --tags` doesn't.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,18 +9,18 @@ IMAGENAME ?= resource-topology-exporter
 IMAGETAG ?= latest
 RTE_CONTAINER_IMAGE ?= quay.io/$(REPOOWNER)/$(IMAGENAME):$(IMAGETAG)
 
-LDFLAGS = -ldflags "-s -w"
-VERSION := $(shell git describe --tags)
-ifneq ($(VERSION),)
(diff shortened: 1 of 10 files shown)


Makefile
```diff
@@ -9,18 +9,18 @@ IMAGENAME ?= resource-topology-exporter
...
 
+.PHONY: build-tools
+build-tools: outdir _out/git-semver
+
 .PHONY: build
-build: outdir
-	$(COMMONENVVAR) $(BUILDENVVAR) go build $(LDFLAGS) -o _out/resource-topology-exporter cmd/resource-topology-exporter/main.go
+build: build-tools
+	$(COMMONENVVAR) $(BUILDENVVAR) go build \
+	-ldflags "-s -w -X github.com/k8stopologyawareschedwg/resource-topology-exporter/pkg/version.version=v$(shell _out/git-semver)" \
```
jlojosnegros, line 22: Shouldn't this 'v' be there?  starting with a prefix this would not be a semver compliant version.

fromanirh: didn't check the spec: this is as close as possible to what the old code did. So let me remove the "v"

PR #92, opened 2021-12-13
e2e: change default ns to all ns

Tal-or:
There are times when we are not fully aware of the ns where the tests are running, so if we want to look for a specific pod (by label) we can look in all namespaces instead.

### DIFF
--- a/test/e2e/utils/testenv/testenv.go
+++ b/test/e2e/utils/testenv/testenv.go
@@ -23,7 +23,7 @@ import (
 const (
 	// we rely on kind for our CI
 	DefaultNodeName             = "kind-worker"
-	DefaultNamespace            = "default"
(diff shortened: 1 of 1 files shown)


fromanirh:
Can't the test runner just override the namespace using the env var? https://github.com/k8stopologyawareschedwg/resource-topology-exporter/blob/master/test/e2e/utils/testenv/testenv.go#L52

Tal-or:
> Can't the test runner just override the namespace using the env var? https://github.com/k8stopologyawareschedwg/resource-topology-exporter/blob/master/test/e2e/utils/testenv/testenv.go#L52

We can. But I thought we want to use the all as the default value.

fromanirh:
I agree this change seems like an improvement over the current state.

Issue #93, opened 2021-12-16
fsnotify deadlock prevention


fromanirh:
per fsnotify docs we should watch for events on a separate goroutine: https://github.com/fsnotify/fsnotify#usage
we need to review if the way we use fsnotify in the `resourcetopologyexporter` pkg is safe and do the necessary changes if it is not.

PR #94, opened 2021-12-17
cleanup concurrency model

fromanirh:
Make sure the goroutines are always handled on the calling sites; streamline all the relevant packages to expose the Run()/Stop() interface.

Resolves: #93

### DIFF
--- a/pkg/resourcetopologyexporter/resourceobserver.go
+++ b/pkg/resourcetopologyexporter/resourceobserver.go
@@ -1,0 +1,77 @@
+package resourcetopologyexporter
+
+import (
+	"fmt"
+	"time"
+
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/klog/v2"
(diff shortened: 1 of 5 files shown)


pkg/resourcetopologyexporter/resourceobserver.go
```diff
@@ -0,0 +1,77 @@
...
+
+			tsBegin := time.Now()
+			monInfo.Zones, err = rm.resMon.Scan(rm.excludeList)
+			tsEnd := time.Now()
+
+			condStatus := v1.ConditionTrue
+			if err != nil {
+				klog.Warningf("failed to scan pod resources: %w\n", err)
+				condStatus = v1.ConditionFalse
+				continue
```
Tal-or, line 66: We have a bug here the condition won't get updated.
But this is just for the record and will handle on a separate PR

PR #95, opened 2021-12-23
Add new ratelimiter to throtlle notifications

jlojosnegros:
Throttle events to avoid storms of podresources pools and
NRT objects updates because both are not cheap operations.

Adapt #85


pkg/ratelimiter/ratelimiter.go
```diff
@@ -0,0 +1,81 @@
+package ratelimiter
```
fromanirh, line 1: do we need the copyright boilerplate?

pkg/ratelimiter/ratelimiter.go
```diff
@@ -0,0 +1,81 @@
...
+
+import (
+	"time"
+
+	"github.com/k8stopologyawareschedwg/resource-topology-exporter/pkg/notification"
+	"go.uber.org/ratelimit"
```
fromanirh, line 7: wrong import order, should read
```
"time"

"go.uber.org/ratelimit"

"github.com/k8stopologyawareschedwg/resource-topology-exporter/pkg/notification"
```

pkg/ratelimiter/ratelimiter.go
```diff
@@ -0,0 +1,81 @@
...
+	rles := RateLimitedEventSource{
+		es:       es,
+		inCh:     es.Events(),
+		outCh:    make(chan notification.Event),
+		doneChan: make(chan struct{}, 1),
+		stopChan: make(chan struct{}),
+	}
+
+	options := ratelimit.Per(timeUnit)
+	rles.rt = ratelimit.New(int(maxEventsPerTimeUnit), options)
```
fromanirh, line 31: let's say we get a burst of 100 events in the time unit, while we are configured to emit only 1 event per time unit. Which event among the initial 100 is going to be emitted? the first (aka the oldest)? the last (aka the newest)? something else?

pkg/ratelimiter/ratelimiter_test.go
```diff
@@ -0,0 +1,133 @@
+package ratelimiter
```
fromanirh, line 1: probably missing copyright boilerplate

pkg/ratelimiter/ratelimiter_test.go
```diff
@@ -0,0 +1,133 @@
...
+	sut.Stop()
+	sut.Wait()
+	done <- struct{}{}
+
+	// Check results
+	if len(results) == 0 {
+		t.Fatalf("Unable to receive results")
+	}
+
+	idealPeriod_us := time.Duration(timeUnit.Microseconds()/numberOfEvents) * time.Microsecond
```
fromanirh, line 94: not a fan of `thisKindOf_mixed_case`

pkg/notification/notification.go
```diff
@@ -24,8 +24,14 @@ type Event struct {
 	Timestamp time.Time
 }
 
-type EventSource struct {
-	Events        <-chan Event
+type EventSource interface {
```
fromanirh, line 27: do we need this interface?

fromanirh (commented):
nice improvements, thanks

pkg/notification/notification.go
jlojosnegros: I have used this interface so the UnlimitedEventSource and the RateLimited one could be handled through it without knowing which one you are using. In fact RateLimitedEventSource just decorates another EventSource adding the rate limitation to it.

pkg/ratelimiter/ratelimiter.go
jlojosnegros: Sorry I did not see it on other files so I forgot it.
I will add it.

pkg/ratelimiter/ratelimiter.go
jlojosnegros: ok. I will change it.,

pkg/notification/notification.go
fromanirh: Sure, but  do we need an explicit declaration? IOW, do we need this level of abstraction yet or can the builder function just return concrete objects?

pkg/ratelimiter/ratelimiter.go
jlojosnegros: None of them. 
I mean, my first approach in #85 was to use a buffered channel as input channel and use a select in the write operation to that channel so when the buffered channel went full the new writes will be dropped, that would make the input and output rates independent. 
But in the review we thought it was better to use a blocking channel. In this way the writer will be blocked in the channel until the reader gets the value, so no burst will occur because the reader will make the writer to go slower.
I still prefer the first approach.

pkg/ratelimiter/ratelimiter_test.go
jlojosnegros: I will add it

pkg/notification/notification.go
jlojosnegros: builder function creates different type of event source from different arguments and should be declared to return only one type so I think the interface is needed, at least to the best of my knowledge.

fromanirh: ok, if nothing else this interface enabled better testing (enables a form of dependency injection) so let's keep it!

fromanirh:
ok, I think this is ready. Please remove the merge commit and squash all the commits in one and we can move forward.

fromanirh (commented):
/lgtm
/approve

PR #97, opened 2022-02-06
e2e: deflake nonintegral cpu request test

Tal-or:
When a guaranteed pod asks for nonintegral CPU request, we expect no changes under the amount of available resources in the NRT object.
This is not completely true and can make the test falky, because the pods gets its memory from the reserved pool (which in turn decreased from the available resources in the NRT object) and then the comparison failed, i.e. initial > final, when the test expect otherwise.

Let's change the test to compare only CPUs, which is eventually what we're mostly care about here.


fromanirh (commented):
/lgtm

swatisehgal:
Thanks @Tal-or for the good catch and submitting the fix!

PR #98, opened 2022-02-28
nrtupdater: report policy changes

fromanirh:
We don't expect frequent topology manager policy changes: actually we deem this to be a very unlikely event, but still
we should update the NRT objects if we detect such a  change, even across a RTE reboot.


fromanirh:
this is a great chance to add a bunch of tests in this area.
Besides that, the PR is reviewable.

swatisehgal (approved):

fromanirh:
thanks @swatisehgal ! I totally want to add tests, here's why the PR is still WIP :)

fromanirh:
test added, the PR is now ready from my POV.

pkg/nrtupdater/nrtupdater.go
```diff
@@ -111,6 +107,12 @@ func (te *NRTUpdater) Update(info MonitorInfo) error {
 	return nil
 }
 
+func (te *NRTUpdater) updateNRTInfo(nrt *v1alpha1.NodeResourceTopology, info MonitorInfo) {
+	nrt.Annotations[AnnotationRTEUpdate] = info.UpdateReason()
+	nrt.TopologyPolicies = []string{te.tmPolicy}
```
Tal-or, line 121: As mentioned before, this data was derived from a command-line argument.
Who's responsible to update that command-line argument properly? the operator?

fromanirh: I don't have a good answer here. It depends on how RTE is deployed. If it is deployed by an operator, then yes, the operator is in charge. However, regardless of this, we need to report the updated value of TM policy should it change for whatever reason. Previously we just didn't do that without no obvious reason to.

swatisehgal: It could be an operator update or manual update in cases where RTE is manually deployed (e.g. update of the command line argument in RTE deployment specification file followed by kubelet apply). Both these scenarios will benefit from this change.

Tal-or: I see the benefit indeed for cases that we can somehow learn Kubelet configuration dynamically (without restarting the pod), but any changes for the command line argument in the RTE deployment will trigger a pod restart, which in turn will update the policy anyway, regardless of this change.

fromanirh: thing is that unless someone ALSO deletes the relevant NRT objects, in case of restart the NRT objects will NOT be updated.

Tal-or: TIL

PR #101, opened 2022-03-23
gh actions: finalize move to go 1.17

fromanirh:
In commit 79d078198d9189bf202c19071dfdf610a40649ff we partially updated to golang 1.17, but we left some gh actions lanes behind to 1.16 for no good reason. Let's fix that.


fromanirh:
known flake

PR #105, opened 2022-04-04
e2e: fix FindNotificationFilePath function

Tal-or:
RTE manifest layout has changed in latest deployer version, when underthe RTE container we splitted the command and arguments of RTE container into separte fileds.
This requires the FindNotificationFilePath to be looking for the --notify-file argument in the correct place.

Furthermore, we didn't check the returned error from the function, let's fix it as well.

### DIFF
--- a/test/e2e/utils/pods/rtepod/rtepod.go
+++ b/test/e2e/utils/pods/rtepod/rtepod.go
@@ -65,13 +65,6 @@ func FindNotificationFilePath(rtePod *corev1.Pod) (string, error) {
 	for idx := 0; idx < len(rtePod.Spec.Containers); idx++ {
 		cnt := rtePod.Spec.Containers[idx] // shortcut
 		if len(cnt.Command) > 0 && strings.Contains(cnt.Command[0], rteExecutable) {
-			for _, arg := range cnt.Command {
-				if strings.Contains(arg, notificationFileOption) {
-					return extractOptionValue(arg)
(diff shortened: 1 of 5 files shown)


test/e2e/utils/pods/rtepod/rtepod.go
```diff
@@ -65,21 +65,14 @@ func FindNotificationFilePath(rtePod *corev1.Pod) (string, error) {
...
-		if len(cnt.Args) > 0 && strings.Contains(cnt.Args[0], rteExecutable) {
 			for _, arg := range cnt.Args {
 				if strings.Contains(arg, notificationFileOption) {
 					return extractOptionValue(arg)
 				}
 			}
 		}
 	}
-	return "", fmt.Errorf("no container uses %q as command or argument option", notificationFileOption)
+	return "", fmt.Errorf("no container uses %q as an argument option", notificationFileOption)
```
fromanirh, line 75: this is a worthy improvement, but I don't think will fix the flakiness, at least not completely.

Tal-or: Why not, the notification file test is always falling at this point. I checked the logs

Tal-or: The whole patch should fix it, not this line specifically

fromanirh:
thanks for the PR!

Issue #84, opened 2021-11-28
create a podresources mock (/proxy?) server for testing


ffromani:
In order to improve the testability, we would like to be able to inject failures and/or delays in the podresources API protocol.
The simplest and most effective way is to create a mock we can control from the test code.
The specifics of the design are to be decided, the bare minimum is to be able to inject failures programmatically.

ffromani:
the mock server (and more) will be implemented in https://github.com/k8stopologyawareschedwg/podresourcesapi-tools

PR #116, opened 2022-07-25
tools: add tool to generate fake load

fromanirh:
With small additions to our existing packages,
we can build a simple NRT load generator, useful to
test, troubleshoot and assess the scalability of RTE.

Note this tool is intentionally not shipped yet in releases:
still needs to mature a bit at least.

### DIFF
--- a/pkg/nrtupdater/fake/fake.go
+++ b/pkg/nrtupdater/fake/fake.go
@@ -1,0 +1,104 @@
+package fake
+
+import (
+	"math/rand"
+	"time"
+
+	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/klog/v2"
+
+	"github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/topology/v1alpha1"
+	"github.com/k8stopologyawareschedwg/resource-topology-exporter/pkg/dump"
+	"github.com/k8stopologyawareschedwg/resource-topology-exporter/pkg/nrtupdater"
+)
+
+type Generator struct {
+	Infos    <-chan nrtupdater.MonitorInfo
+	infoChan chan nrtupdater.MonitorInfo
+	interval time.Duration
+}
+
+func NewGenerator(interval time.Duration) *Generator {
+	gen := Generator{
+		infoChan: make(chan nrtupdater.MonitorInfo),
+		interval: interval,
+	}
+	gen.Infos = gen.infoChan
+	return &gen
+}
+
+func (ge *Generator) Run() {
+	ticker := time.NewTicker(ge.interval)
+
+	for {
+		select {
+		case <-ticker.C:
+			mi := nrtupdater.MonitorInfo{
+				Timer: true,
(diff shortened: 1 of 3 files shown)


pkg/nrtupdater/fake/fake.go
```diff
@@ -0,0 +1,106 @@
...
+	ticker := time.NewTicker(ge.interval)
+	anns := make(map[string]string)
+
+	for {
+		select {
+		case <-ticker.C:
+			mi := nrtupdater.MonitorInfo{
+				Timer:       true,
+				Zones:       Zones(),
+				Annotations: anns,
```
Tal-or, line 40: I don't see it's being used somewhere, why do we need to initialize that?

tools/nrtstress/main.go
```diff
@@ -0,0 +1,52 @@
...
+)
+
+func main() {
+	var hostname string
+	var tmPolicy string
+	var interval time.Duration
+	var seed int
+	var dryRun bool
+	flag.StringVar(&hostname, "hosthame", "fake-host-0", "fake host name (not validated)")
+	flag.StringVar(&tmPolicy, "tm-policy", "best-effort", "topology manager policy (not validated)")
```
Tal-or, line 21: IMHO we should set the default to single-numa-node, in case we would like to use this generator together with the scheduler, we won't extra configuration here.

tools/nrtstress/main.go
```diff
@@ -0,0 +1,52 @@
...
+
+	klog.Infof("generating fake periodic updates every %v", interval)
+
+	gen := fake.NewGenerator(interval)
+	go gen.Run()
+
+	klog.Infof("using NRT Updater args: %+#v", nrtupdaterArgs)
+
+	upd := nrtupdater.NewNRTUpdater(nrtupdaterArgs, tmPolicy)
+	upd.Run(gen.Infos, nil)
```
Tal-or, line 51: I would expect the generator to send the data to some 'real' updater which is part of the existing stack.
I don't understand how sending fake data to a fake updater can be helpful. 
What am I missing here?

pkg/nrtupdater/fake/fake.go
fromanirh: good point, started more ambitious and then downsized. We don't need annotations yet.

tools/nrtstress/main.go
fromanirh: fair point.

tools/nrtstress/main.go
fromanirh: I should have given more context in the commit message/PR cover letter, I'll fix that.
This tool want to stress the apiserver part; we can easily generate fake "realistic" NRT data, so we can easily assess the apiserver load. Generating fake data from the kubelet side requires more effort (https://github.com/k8stopologyawareschedwg/resource-topology-exporter/issues/84)

Tal-or:
/lgtm
/approve
Thank you for the changes

PR #120, opened 2022-08-25
Enable dynamic resource creation option

Tal-or:
Creating new resources (usually devices) doesn't reflect under the NRT object.
The logic to support was already there, but the option to enable it was never exposed to the user. 

For the start, this option would be disabled by default until we'll be sure it's stable enough 

- [x] expose the flag
- [x] fix capacity
- [x] unit-test
- [x] e2e test
- [x] decide whether we want to enable this option by default
- [x] consider decoupling this logic from the standard polling routine.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -73,6 +73,12 @@ jobs:
         kubectl logs -l name=resource-topology -c resource-topology-exporter-container || :
         kubectl get noderesourcetopologies.topology.node.k8s.io -A -o yaml
 
+    - name: deploy sample-devices
+      run: | 
+        hack/deploy-devices.sh
+        hack/check-ds.sh default device-plugin-a-ds
+        kubectl describe nodes -l node-role.kubernetes.io/worker= || :
+
     - name: run E2E tests
       run: |
         export KUBECONFIG=${HOME}/.kube/config 
--- a/pkg/resourcemonitor/resourcemonitor.go
+++ b/pkg/resourcemonitor/resourcemonitor.go
@@ -19,6 +19,8 @@ package resourcemonitor
 import (
 	"context"
 	"fmt"
+	"reflect"
+
 	"os"
 	"strconv"
 	"strings"
@@ -27,11 +29,14 @@ import (
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog/v2"
 	podresourcesapi "k8s.io/kubelet/pkg/apis/podresources/v1"
 
 	"github.com/jaypipes/ghw"
 	"github.com/k8stopologyawareschedwg/podfingerprint"
+	"github.com/k8stopologyawareschedwg/resource-topology-exporter/pkg/k8shelpers"
 
 	topologyv1alpha1 "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/topology/v1alpha1"
 	"github.com/k8stopologyawareschedwg/resource-topology-exporter/pkg/prometheus"
@@ -112,17 +117,20 @@ func NewResourceMonitorWithTopology(nodeName string, topo *ghw.TopologyInfo, pod
 		topo:              topo,
 		coreIDToNodeIDMap: MakeCoreIDToNodeIDMap(topo),
 	}
+
 	if !rm.args.RefreshNodeResources {
 		klog.Infof("getting node resources once")
-		if err := rm.updateNodeCapacity(); err != nil {
+		if err := rm.updateNodeResources(); err != nil {
 			return nil, err
 		}
-		if err := rm.updateNodeAllocatable(); err != nil {
-			return nil, err
-		}
-		rm.updateDevicesCapacity()
 	} else {
-		klog.Infof("getting allocatable resources before each poll")
+		klog.Infof("tracking node resources")
+		if err := rm.updateNodeResources(); err != nil {
+			return nil, err
+		}
+		if err := addNodeInformerEvent(cache.ResourceEventHandlerFuncs{UpdateFunc: rm.resUpdated}); err != nil {
+			return nil, err
+		}
 	}
 
 	if rm.args.Namespace != "" {
@@ -134,16 +142,6 @@ func NewResourceMonitorWithTopology(nodeName string, topo *ghw.TopologyInfo, pod
 }
 
 func (rm *resourceMonitor) Scan(excludeList ResourceExcludeList) (topologyv1alpha1.ZoneList, map[string]string, error) {
-	if rm.args.RefreshNodeResources {
-		if err := rm.updateNodeCapacity(); err != nil {
-			return nil, nil, err
-		}
-		if err := rm.updateNodeAllocatable(); err != nil {
-			return nil, nil, err
-		}
-		rm.updateDevicesCapacity()
-	}
-
 	ctx, cancel := context.WithTimeout(context.Background(), defaultPodResourcesTimeout)
 	defer cancel()
 	resp, err := rm.podResCli.List(ctx, &podresourcesapi.ListPodResourcesRequest{})
@@ -259,6 +257,28 @@ func (rm *resourceMonitor) updateNodeCapacity() error {
 	return nil
 }
 
+func (rm *resourceMonitor) resUpdated(old, new interface{}) {
+	nOld := old.(*v1.Node)
+	nNew := new.(*v1.Node)
+
+	if nNew.Name != rm.nodeName {
+		return
+	}
+
+	// the status frequency update are configurable via the node-status-update-frequency option in Kubelet
+	if !reflect.DeepEqual(nOld.Status.Capacity, nNew.Status.Capacity) ||
+		!reflect.DeepEqual(nOld.Status.Allocatable, nNew.Status.Allocatable) {
+		klog.V(4).Infof("update node resources")
+		if err := rm.updateNodeCapacity(); err != nil {
+			klog.ErrorS(err, "error while updating node capacity")
+		}
+		if err := rm.updateNodeAllocatable(); err != nil {
+			klog.ErrorS(err, "error while updating node allocatable")
+		}
+		rm.updateDevicesCapacity()
+	}
+}
+
 func (rm *resourceMonitor) updateNodeAllocatable() error {
 	ctx, cancel := context.WithTimeout(context.Background(), defaultPodResourcesTimeout)
 	defer cancel()
@@ -285,6 +305,17 @@ func (rm *resourceMonitor) updateDevicesCapacity() {
 			capacityResCnt[resName] = quan
 		}
 	}
+}
+
+func (rm *resourceMonitor) updateNodeResources() error {
+	if err := rm.updateNodeCapacity(); err != nil {
+		return err
+	}
+	if err := rm.updateNodeAllocatable(); err != nil {
+		return err
+	}
+	rm.updateDevicesCapacity()
+	return nil
 }
 
 func GetAllContainerDevices(podRes []*podresourcesapi.PodResources, namespace string, coreIDToNodeIDMap map[int]int) []*podresourcesapi.ContainerDevices {
@@ -353,7 +384,6 @@ func NormalizeContainerDevices(devices []*podresourcesapi.ContainerDevices, memo
 			})
 		}
 	}
-
 	return contDevs
 }
 
@@ -444,4 +474,21 @@ func cpuCapacity(topo *ghw.TopologyInfo, nodeID int) int64 {
 		logicalCoresPerNUMA += len(core.LogicalProcessors)
 	}
 	return int64(logicalCoresPerNUMA)
+}
+
+func addNodeInformerEvent(handler cache.ResourceEventHandlerFuncs) error {
+	c, err := k8shelpers.GetK8sClient("")
+	if err != nil {
+		return err
+	}
+	// node-status-update-frequency
+	factory := informers.NewSharedInformerFactory(c, 0)
+	nodeInformer := factory.Core().V1().Nodes().Informer()
+	nodeInformer.AddEventHandler(handler)
+	ctx := context.Background()
+	factory.Start(ctx.Done())
+	if !cache.WaitForCacheSync(ctx.Done(), nodeInformer.HasSynced) {
+		return fmt.Errorf("timed out waiting for caches to sync")
+	}
+	return nil
 }
--- a/pkg/resourcemonitor/resourcemonitor.go
+++ b/pkg/resourcemonitor/resourcemonitor.go
@@ -120,6 +120,7 @@ func NewResourceMonitorWithTopology(nodeName string, topo *ghw.TopologyInfo, pod
 		if err := rm.updateNodeAllocatable(); err != nil {
 			return nil, err
 		}
+		rm.updateDevicesCapacity()
 	} else {
 		klog.Infof("getting allocatable resources before each poll")
 	}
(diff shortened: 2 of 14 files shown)


swatisehgal:
Looks good to me so far. I will review this PR again once we have added some tests and addressed the other bullet points.

pkg/resourcemonitor/resourcemonitor.go
```diff
@@ -271,6 +273,20 @@ func (rm *resourceMonitor) updateNodeAllocatable() error {
 	return nil
 }
 
+func (rm *resourceMonitor) updateDevicesCapacity() {
```
fromanirh, line 327: for consistency and logical ordering, we should probably update the capacity, then the allocatable. Not too big of a deal, though.

pkg/resourcemonitor/resourcemonitor.go
```diff
@@ -259,6 +257,28 @@ func (rm *resourceMonitor) updateNodeCapacity() error {
 	return nil
 }
 
+func (rm *resourceMonitor) resUpdated(old, new interface{}) {
```
fromanirh, line 291: at first I was a bit skeptical towards this approach, but the more I read about it the more I like it. Good idea, let's try this.

pkg/resourcemonitor/resourcemonitor.go
```diff
@@ -259,6 +257,28 @@ func (rm *resourceMonitor) updateNodeCapacity() error {
...
+	nNew := new.(*v1.Node)
+
+	if nNew.Name != rm.nodeName {
+		return
+	}
+
+	// the status frequency update are configurable via the node-status-update-frequency option in Kubelet
+	if !reflect.DeepEqual(nOld.Status.Capacity, nNew.Status.Capacity) ||
+		!reflect.DeepEqual(nOld.Status.Allocatable, nNew.Status.Allocatable) {
+		klog.V(4).Infof("update node resources")
```
fromanirh, line 302: if this should happen "rarely", we can use `V(3)` or even `V(2)`

pkg/resourcemonitor/resourcemonitor.go
```diff
@@ -445,3 +475,20 @@ func cpuCapacity(topo *ghw.TopologyInfo, nodeID int) int64 {
...
+	c, err := k8shelpers.GetK8sClient("")
+	if err != nil {
+		return err
+	}
+	// node-status-update-frequency
+	factory := informers.NewSharedInformerFactory(c, 0)
+	nodeInformer := factory.Core().V1().Nodes().Informer()
+	nodeInformer.AddEventHandler(handler)
+	ctx := context.Background()
+	factory.Start(ctx.Done())
```
fromanirh, line 515: do we want to use all the informer/factory object as throwaways or do we want/need to store references to them in the main ResourceMonitor object?

pkg/resourcemonitor/resourcemonitor.go
```diff
@@ -19,6 +19,7 @@ package resourcemonitor
 import (
 	"context"
 	"fmt"
+	"github.com/k8stopologyawareschedwg/resource-topology-exporter/pkg/k8shelpers"
```
fromanirh, line 22: since we're moving imports, this belongs in the group below alongside podsfingerprint

pkg/resourcemonitor/resourcemonitor.go
```diff
@@ -92,30 +92,34 @@ type resourceMonitor struct {
...
-		"/sys": args.SysfsRoot,
-	}))
-	if err != nil {
-		return nil, err
+func NewResourceMonitor(podResCli podresourcesapi.PodResourcesListerClient, args Args, options ...func(*resourceMonitor)) (*resourceMonitor, error) {
+	rm := &resourceMonitor{
+		podResCli: podResCli,
+		args:      args,
+	}
+	for _, o := range options {
```
fromanirh, line 107: I for one actually like the single-letter identifiers, part of the UNIX heritage and actually good enough in very limited scope like this. But they also turned hard to grep, so let's try something a bit longer:
```
for _, opt := range options {
   ...
```

pkg/resourcemonitor/resourcemonitor.go
```diff
@@ -92,30 +92,34 @@ type resourceMonitor struct {
...
-		topo:              topo,
-		coreIDToNodeIDMap: MakeCoreIDToNodeIDMap(topo),
+	if rm.topo == nil {
+		topo, err := ghw.Topology(ghw.WithPathOverrides(ghw.PathOverrides{
+			"/sys": args.SysfsRoot,
+		}))
+		if err != nil {
+			return nil, err
+		}
+		rm.topo = topo
```
fromanirh, line 118: missing
```
rm.coreIDToNodeIDMap = MakeCoreIDToNodeIDMap(topo)
```
?

fromanirh (commented):
overall very nice, I like the general improvements introduced here.

pkg/resourcemonitor/resourcemonitor.go
Tal-or: I think that is the bug I'm struggling with :)

pkg/resourcemonitor/resourcemonitor.go
Tal-or: Besides adding the event, I didn't see a reason to keep a reference for it.
We can if we think we would have a future use for it. but atm I don't see one

fromanirh: I didn't look hard/deep enough, I'm mostly wondering about side effects here. Do we _need_ to keep reference for the code to work? AFAIR no, asking because I didn't look hard/deep enough yet :)

pkg/resourcemonitor/resourcemonitor.go
Tal-or: We are relying on the allocatable value for updating the capacity since we do not have a trivial way to detect the capacity directly.

pkg/resourcemonitor/resourcemonitor.go
Tal-or: I had exactly the same question but I tested it with and without keeping the reference and it worked.
I guess the reason is that it's a shared informer and when we start it, it starts a go routine to keep track of the events that keep the object alive.
At least this is how I understood.

fromanirh: ok, good enough for now then!

pkg/resourcemonitor/resourcemonitor.go
fromanirh: ok, this is however a bit counter-intuitive so it deserves a code comment (perhaps bouncing back to the commit message/the place on which we spend time explaining why)

Tal-or: Is the comment in lines 283-284 isn't enough?

fromanirh: Contentwise - yes. But is buried very deep in the function, we should probably move it (better move than copy in this case IMO) upper in the stack. Perhaps in the call site?

.github/workflows/e2e.yml
```diff
@@ -73,6 +73,12 @@ jobs:
         kubectl logs -l name=resource-topology -c resource-topology-exporter-container || :
         kubectl get noderesourcetopologies.topology.node.k8s.io -A -o yaml
 
+    - name: deploy sample-devices
+      run: | 
+        hack/deploy-devices.sh
+        hack/check-ds.sh default device-plugin-a-ds
+        kubectl describe nodes -l node-role.kubernetes.io/worker= || :
```
fromanirh, line 80: this means we create the devices before RTE starts, so when RTE starts it has the devices. I'm not sure this is the scenario we want to check - or, at least, that's not the ONLY scenario we want to check. We should also check the scenario on which devices are created after RTE started.

fromanirh (commented):
let's clarify the e2e tests, everything else LGTM

.github/workflows/e2e.yml
Tal-or: That's exactly the scenario we have here.
Initially, we are running the RTE pods (in ```deploy RTE```), waiting for it to be ready (in ```cluster-ready```) and only then deploy the devices (in ```deploy sample-devices``)

Tal-or:
There is a time when we're having the initial scan on the main thread but the informer is not starting yet.
That could lead to an event (device creation) that will occur just between those two actions and then won't be reflected on the NRT object.
This is a **very unlikely** scenario, but it would be almost impossible for detecting on user's environment and that's something that we would like to avoid,

.github/workflows/e2e.yml
fromanirh: uhm, ok. This is good enough for a start

fromanirh (approved):
LGTM
let's give time to @swatisehgal to review

pkg/resourcemonitor/resourcemonitor.go
```diff
@@ -259,6 +257,28 @@ func (rm *resourceMonitor) updateNodeCapacity() error {
...
+	if !reflect.DeepEqual(nOld.Status.Capacity, nNew.Status.Capacity) ||
+		!reflect.DeepEqual(nOld.Status.Allocatable, nNew.Status.Allocatable) {
+		klog.V(4).Infof("update node resources")
+		if err := rm.updateNodeCapacity(); err != nil {
+			klog.ErrorS(err, "error while updating node capacity")
+		}
+		if err := rm.updateNodeAllocatable(); err != nil {
+			klog.ErrorS(err, "error while updating node allocatable")
+		}
+		rm.updateDevicesCapacity()
```
swatisehgal, line 309: We can simply call `updateNodeResources` here instead of reimplementing that logic here.

test/e2e/utils/testenv/testenv.go
```diff
@@ -72,6 +73,13 @@ func GetPollInterval() string {
 	return pollInterval
 }
 
+func GetDeviceName() string {
+	if devName, ok := os.LookupEnv("E2E_DEVICE_NAME"); ok {
```
swatisehgal, line 77: IIUC, this should be `DEVICE_RESOURCE_NAME`. We are creating environment variable when deploying the device plugin [here](https://github.com/k8stopologyawareschedwg/resource-topology-exporter/pull/120/commits/91f7d2059d1c92acebfde4f1e3829c2618050eff#diff-e19ff91c1b014f024d0a56149f75ceb456dbb075243fe8971143611def27140bR20) where the key is  `DEVICE_RESOURCE_NAME`. Is `E2E_DEVICE_NAME` meant to be a separate environment variable that the we are supposed to create? Can you please clarify if I am missing something here?

pkg/resourcemonitor/resourcemonitor.go
swatisehgal: +1 on this.

swatisehgal (commented):
Overall the PR is looking good. Thanks for all the work on this.

Left a few comments.

swatisehgal:
> > everything else LGTM
> 
> There is a time when we're having the initial scan on the main thread but the informer is not starting yet. That could lead to an event (device creation) that will occur just between those two actions and then won't be reflected on the NRT object. This is a **very unlikely** scenario, but it would be almost impossible for detecting on user's environment and that's something that we would like to avoid.

Will think about this a little bit more about this scenario and provide my comments.

pkg/resourcemonitor/resourcemonitor.go
fromanirh: good catch, this seems like a good idea

Tal-or: The reason why I didn't do that is that we can't return the error from ```updateNodeResources```  and it would be more clear if we would be able to log explicitly from which of the functions the error came from.

test/e2e/utils/testenv/testenv.go
Tal-or: The ```DEVICE_RESOURCE_NAME``` is defined inside the device-plugin pod and we don't have an access to it in the test scope.
We need the ```E2E_DEVICE_NAME``` so we could tell the test the exact name to look for in case we would import the new test in other packages.
Like we are doing here:
https://github.com/openshift-kni/numaresources-operator/blob/main/test/e2e/rte/rte_e2e_suite_test.go#L32

pkg/resourcemonitor/resourcemonitor.go
fromanirh: what about:
```
func (rm *resourceMonitor) updateNodeResources() error {
	if err := rm.updateNodeCapacity(); err != nil {
		return fmt.Errorf("error while updating node capacity: %w", err)
	}
	if err := rm.updateNodeAllocatable(); err != nil {
		return fmt.Errorf("error while updating node allocatable: %w", err)
	}
	rm.updateDevicesCapacity()
	return nil
}

...
	klog.V(3).Infof("update node resources")
	if err := rm.updateNodeResources(); err != nil {
		klog.ErrorS(err, "while updating node resources")
	}
...

```

Tal-or: Nice!

Tal-or:
> > > everything else LGTM
> > 
> > 
> > There is a time when we're having the initial scan on the main thread but the informer is not starting yet. That could lead to an event (device creation) that will occur just between those two actions and then won't be reflected on the NRT object. This is a **very unlikely** scenario, but it would be almost impossible for detecting on user's environment and that's something that we would like to avoid.
> 
> Will think about this scenario a little bit more and provide my comments.

I'm concern handling this edge case will require a lock, or at least moving the update logic back to the scan loop

fromanirh:
> > > > everything else LGTM
> > > 
> > > 
> > > There is a time when we're having the initial scan on the main thread but the informer is not starting yet. That could lead to an event (device creation) that will occur just between those two actions and then won't be reflected on the NRT object. This is a **very unlikely** scenario, but it would be almost impossible for detecting on user's environment and that's something that we would like to avoid.
> > 
> > 
> > Will think about this scenario a little bit more and provide my comments.
> 
> I'm concerned handling this edge case will require a lock, or at least moving the update logic back to the scan loop

this is a legit concern which can be handled in a followup patch.
The very first thing coming to mind (so debatable) is that we should have only a goroutine handling the scans. If the main thread wants to trigger a scan (= initial scan), it can just send a request into a channel, which will be serviced ASAP.

test/e2e/utils/testenv/testenv.go
swatisehgal: Ack, thanks for clarifying this.

pkg/resourcemonitor/resourcemonitor.go
swatisehgal:
```
what about:

func (rm *resourceMonitor) updateNodeResources() error {
	if err := rm.updateNodeCapacity(); err != nil {
		return fmt.Errorf("error while updating node capacity: %w", err)
	}
	if err := rm.updateNodeAllocatable(); err != nil {
		return fmt.Errorf("error while updating node allocatable: %w", err)
	}
	rm.updateDevicesCapacity()
	return nil
}

...
	klog.V(3).Infof("update node resources")
	if err := rm.updateNodeResources(); err != nil {
		klog.ErrorS(err, "while updating node resources")
	}
...
```

Exactly what I was trying to suggest. Thanks @fromanirh.

Tal-or:
> this is a legit concern which can be handled in a followup patch.

Ack. Then I'll open an issue to track it.

> The very first thing coming to mind (so debatable) is that we should have only a goroutine handling the scans. If the main thread wants to trigger a scan (= initial scan), it can just send a request into a channel, which will be serviced ASAP.

This is possible, it'll requires the scan loop to have additional channel in the ```select``` but that's OK I guess.

fromanirh:
> > The very first thing coming to mind (so debatable) is that we should have only a goroutine handling the scans. If the main thread wants to trigger a scan (= initial scan), it can just send a request into a channel, which will be serviced ASAP.
> 
> This is possible, it'll requires the scan loop to have additional channel in the `select` but that's OK I guess.

Maybe not. We can possibly generate an event at startup time using the existing event source.

Tal-or:
> > > The very first thing coming to mind (so debatable) is that we should have only a goroutine handling the scans. If the main thread wants to trigger a scan (= initial scan), it can just send a request into a channel, which will be serviced ASAP.
> > 
> > 
> > This is possible, it'll requires the scan loop to have additional channel in the `select` but that's OK I guess.
> 
> Maybe not. We can possibly generate an event at startup time using the existing event source.

The Scan loop doesn't contain the update logic. if we would add it to the existing event source, we again, coupling the capacity and allocatable update logic with the available update logic.

That means we should have a separate channel or at least a different type of event.

fromanirh (approved):
LGTM

PR #122, opened 2022-09-05
config: introduce the configuration package

fromanirh:
Move the config logic, currently handling the configuration flags, into a proper package. Benefits of this move:
1. better testability
2. enable code reuse
3. enable future refactorings/changes
4. open the opportunity for code simplification
5. allows us to take the chance to rewrite the unit tests, shedding a dep - we should really use the simple testing package here unless  something more powerful is clearly needed.

PR #123, opened 2022-09-06
config: resmon: simplify excludelist

fromanirh:
Remove the nested excludelist structure: a simpler type alias is sufficient. Besides the internal API change, no intended change of behavior. The tests will ensure so.

Depends on #122


pkg/config/config.go
```diff
@@ -105,6 +110,13 @@ func LoadArgs(args ...string) (ProgArgs, error) {
 	pArgs.RTE.TimeUnitToLimitEvents = time.Second
 
 	flags.BoolVar(&pArgs.Version, "version", false, "Output version and exit")
+	flags.StringVar(&pArgs.DumpConfig, "dump-config", "",
```
swatisehgal, line 113: In case the user hasn't specified a file path, we are not providing a default dump-config file path and choosing to not dump the config. This is also applicable if dump-config is explicitly specified to be an empty string. Might be worth mentioning this in the flag description so that the user does not assume that a default file path would be picked up and the config would still be dumped.

fromanirh: that's a good point. Will change in the next upload.

Tal-or:
LGTM

PR #127, opened 2022-09-08
resmon: reduce spam in the logs with NUMA affinity

fromanirh:
When a device plugin doesn't report its NUMA affinity, we are very likely to get log spammed with
```
W0903 02:32:37.402438       1 resourcemonitor.go:198] zero capacity for resource "example.com/deviceA" on NUMA cell 0
W0903 02:32:37.402438       1 resourcemonitor.go:198] zero capacity for resource "example.com/deviceA" on NUMA cell 1
```

Log spamming is not just annoying: it causes log rotation and hides useful information.
Let's take steps to improve this state, trying to avoid to just remove a potentially useful log message.


swatisehgal (approved):
/lgtm

Thanks for the PR. This is a great catch!

PR #134, opened 2022-10-17
pfp: debug: add option to debug the pfp status

fromanirh:
pfp >= v0.1.0 is gaining debug aids. We add an option to track and dump the pfp status (= podset considered, pfp sign emitted) after each computation, overwriting the previous status.

Caveats:
- the write is SYNCHRONOUS in Scan(). This has both pros and cons, but we start like this mostly because it's the simplest approach.
- the option is disabled by default.


fromanirh:
the failure is a known flake (which we should fix eventually)

PR #137, opened 2022-11-03
deps: k8s 1.23.3 -> 1.24.7

fromanirh:
trivial dep bump

Signed-off-by: Francesco Romani <fromani@redhat.com>

PR #143, opened 2022-11-04
deps: k8s: 1.24.7 -> 1.25.3

fromanirh:
trivial dep bump, but with a tail of updates triggered by the fact kube 1.25+ now requires ginkgo v2


fromanirh:
CI failure is real and we need updates to our e2e tests

fromanirh:
kube 1.25 pulls ginkgo v1. This means that before being able to update to kube 1.25.z, we need to port our tests to ginkgo v2 (or alongside)

swatisehgal (approved):
/lgtm

When I reviewed https://github.com/k8stopologyawareschedwg/resource-topology-exporter/pull/137 yesterday, I was wondering why we were not moving to the latest version but then thought that we were intentionally moving to the next version :) 

Anyway, I am definitely in the favor of moving the latest version! Thanks for taking care of this work.

fromanirh:
> /lgtm
> 
> When I reviewed #137 yesterday, I was wondering why we were not moving to the latest version but then thought that we were intentionally moving to the next version :)
> 
> Anyway, I am definitely in the favor of moving the latest version! Thanks for taking care of this work.

That's a very good point. I wanted to have the intermediate step recorded (release-0.8), that's the main driver

PR #151, opened 2023-01-11
e2e: try again to simplify/deflake smart poller tests

ffromani:
Another attempt to simplify and at the same time deflake the longstanding unreliable smart poller test.

### DIFF
--- a/test/e2e/rte/rte.go
+++ b/test/e2e/rte/rte.go
@@ -110,52 +110,36 @@ var _ = ginkgo.Describe("[RTE][InfraConsuming] Resource topology exporter", func
 			ginkgo.By("creating a pod consuming the shared pool")
 			sleeperPod := e2epods.MakeGuaranteedSleeperPod("1000m")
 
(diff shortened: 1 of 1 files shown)


Tal-or:
I would suggest to cancel the update by interval in this test. (we can set `sleepInterval = 0` to imply no update on interval basis)
This way we can guaranteed that when an update received it received for the right reasons, and we don't need to schedule the pod creation between intervals.

ffromani:
the work to disable polling, which requires quite a lot of infra reworking, is underway in #153

swatisehgal (approved):
I see there is work in progress to tackle cases where we tackle cases for polling and evented separately for more focused testing. For now, I think we can proceed and merge this.

 /lgtm

ffromani:
> I see there is work in progress to tackle cases where we tackle cases for polling and evented separately for more focused testing. For now, I think we can proceed and merge this.
> 
> /lgtm

yes, the two efforts complement each other. Thanks!

PR #152, opened 2023-01-11
notification: allow to disable the polling

ffromani:
Up until now, it was not possible to disable the polling. The workaround was to set the polling interval to a insanely high value (like 1 day) but still, this is not a real fix.

The assumption was some form of polling was always useful, but over time this turns out to be not really true, and disabling polling is helpful at very least in testing.

Implying some form of polling was more than a historical combination of events than a conscious design decision, so let's fix it, better late than never.

### DIFF
--- a/pkg/notification/notification.go
+++ b/pkg/notification/notification.go
@@ -44,17 +44,16 @@ type UnlimitedEventSource struct {
 	doneChan      chan struct{}
 }
 
-func NewUnlimitedEventSource(sleepInterval time.Duration) (*UnlimitedEventSource, error) {
+func NewUnlimitedEventSource() (*UnlimitedEventSource, error) {
 	watcher, err := fsnotify.NewWatcher()
 	if err != nil {
 		return nil, fmt.Errorf("failed to create the watcher: %w", err)
 	}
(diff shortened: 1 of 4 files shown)


pkg/notification/notification.go
```diff
@@ -111,6 +118,18 @@ func (es *UnlimitedEventSource) Run() {
 	es.doneChan <- struct{}{}
 }
 
+func (es *UnlimitedEventSource) AddInterval(interval time.Duration) error {
```
Tal-or, line 121: Basically this function can be read as `DisableInterval`
any other option return an error :)

ffromani: Enables the interval just once, and only if the value is >= 0 :)

ffromani: I'm torn if this should be renamed `SetInterval` to emphasize that we can have 0-1 timers in the current implementation :\

Tal-or: > I'm torn if this should be renamed SetInterval to emphasize that we can have 0-1 timers in the current implementation :\

Sounds more accurate than `AddInterval`

PR #153, opened 2023-01-12
WIP: e2e: split lanes

ffromani:
split the e2e lane in polling and evented, with different RTE configurations: in the former, use only polling; in the latter,
use only events. This is meant to enable more focused, thus precise, e2e tests.

A lane with both enabled is intentionally left out for now, given the high CI cost is unclear the benefits it will give us.

Some tests which are not related to the update mechanism are intentionally run in both scenarios. The highest cost is the
complilation of code, which is shared among lanes.


Tal-or:
@ffromani Is this work is still in progress? if not let's reduce the noise and close it

ffromani:
> @ffromani Is this work is still in progress? if not let's reduce the noise and close it

Still WIP. Will keep it open for a little while more.

ffromani:
need to restart from scratch when we have more headroom

PR #154, opened 2023-01-16
e2e: deflake e2e pfp tests

ffromani:
we have longstanding PFP flakes. The root cause is we asyncronously delete test pods, so the pod state leaks from a test to another.
This is because we delete test pod asynchronously in defer(), which leads to leak
A more robust approach is to synchronously delete pods inside ginkgo v2 `DeferCleanup`.
As extra bonus, we can now get rid of the `Cooldown` hack.

Found while testing/tuning #153


swatisehgal (approved):
/lgtm

Thanks for your work on this!

Issue #150, opened 2023-01-02
add label with version and git commit


ffromani:
add labels to the container image to hold the version and the git commit from which the binary was built

jlojosnegros:
I could give it a try, if that is ok

ffromani:
> I could give it a try, if that is ok

go for it!

PR #155, opened 2023-01-19
Add labels to container image: version and commit

jlojosnegros:
Add labels to the container image to hold the version and the git commit from which the binary was built.
Use opencontainer annotations defined
[here](https://github.com/opencontainers/image-spec/blob/main/annotations.md)

Issue: #150


ffromani (approved):
LGTM, thanks!

PR #156, opened 2023-01-30
nrt: expose TM config as attributes

ffromani:
expose TM configuration as attributes of a special Zone. This is in addition to TopologyPolicies field, which is being deprecated. See:
- https://github.com/k8stopologyawareschedwg/noderesourcetopology-api/issues/24
- https://github.com/k8stopologyawareschedwg/noderesourcetopology-api/pull/25

### DIFF
--- a/pkg/nrtupdater/nrtupdater.go
+++ b/pkg/nrtupdater/nrtupdater.go
@@ -24,6 +24,9 @@ const (
 	RTEUpdatePeriodic = "periodic"
 	RTEUpdateReactive = "reactive"
 )
+const (
+	ZoneConfigType = "TopologyManager"
+)
 
 // Command line arguments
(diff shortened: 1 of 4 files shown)


ffromani:
cc @jlojosnegros

swatisehgal (commented):
I would be in the favor of getting the `v1alpha2` version of the API merged first and updating the top level attributes the right way rather than using a dummy zone for populating this information. Captured my opinion and the reason behind that on the corresponding [PR](https://github.com/kubernetes-sigs/scheduler-plugins/pull/488#issuecomment-1409044854) proposed in the scheduler plugin repo as well.

ffromani:
/hold
let's wait and update then!

ffromani:
replaced by #157

PR #161, opened 2023-02-07
notif: deprecate --kubelet-state-dir

ffromani:
The file-based notification is easier to secure and more reliable, because the state-dir approach handles deletion events with unavoidable delay - this is just how kubelet operates, we can't fix this.

### DIFF
--- a/pkg/notification/notification.go
+++ b/pkg/notification/notification.go
@@ -162,6 +162,10 @@ func (es *UnlimitedEventSource) AddDirs(kubeletStateDirs []string) error {
 		return nil
 	}
 
+	klog.Infof("**DEPRECATED** watching state directories is insecure and has known issues")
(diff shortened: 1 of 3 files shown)


pkg/notification/notification.go
```diff
@@ -162,6 +162,10 @@ func (es *UnlimitedEventSource) AddDirs(kubeletStateDirs []string) error {
 		return nil
 	}
 
+	klog.Infof("**DEPRECATED** watching state directories is insecure and has known issues")
+	klog.Infof("**DEPRECATED** watching state directories eill be removed in a future version")
```
MarSik, line 166:
```suggestion
	klog.Infof("**DEPRECATED** watching state directories will be removed in a future version")
```

ffromani: thanks, fixed

swatisehgal (approved):
/lgtm
/approve

PR #162, opened 2023-03-16
rte: params: fix misdetection of the TM params

ffromani:
In case the topology manager parameters where detected from the kubelet config, the attributes where incorrectly populated with the (empty) values from the command line, effectively discarded them.

The topologyPolicies field was however correctly populated.


ffromani:
known flake

PR #167, opened 2023-03-30
modularize and augment the podrescli package

ffromani:
the `podrescli` package was a middleware in disguise. Let's make it official and let's organize the podres utilities in a proper hierarchy, adding middleware and filters.


pkg/podres/middleware/excludeset/excludeset.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+	return &filteringClient{
+		debug:       debug,
+		cli:         cli,
+		podExcludes: podExcludes,
+	}
+}
+
+func ShouldExclude(podExcludes map[string]string, namespace, name string, debug bool) bool {
+	for namespaceGlob, nameGlob := range podExcludes {
```
Tal-or, line 81: Can't we just look for namespace and name of the pod instead of iterating the whole map?
Something like:
```
if v, ok := podExcludes[namespace]; !ok {
    return
}
```
About the names I didn't understood the mapping because for namespace key there could be multiple names with the same key i.e multiple pods in the same namespace.
So how do you exactly handle it inside a map?

ffromani: this comment is yet to be addressed.

PR #174, opened 2023-05-03
pfp: expose computation method

ffromani:
recently we gained a way to compute the PFP directly, considering all pods (standard behavior up until now) or only pods with exclusive resources allocated to them.

Expose the computation method as NRT attribute, to enable the consumers to know explicitely how the PFP was computed and react accordingly.


Tal-or:
/lgtm
/approve
I'm not an expert on that area, but look like a trivial change armed with an e2e test + all test are passing so LGTM.

PR #178, opened 2023-05-23
resmon: pfp: expose pfp compute method

ffromani:
The previously-added `--pods-fingerprint-unrestricted` was clumsy and unexpressive. Drop it and add a more generic and expressive `--pods-fingerprint-method` which accepts (potentially a subset of) the same methods exposed by the podfingerprint package.

While the input interface obviously change, the produced output is expected to be identical.

This change should not be backported to the v0.11 branch.


swatisehgal (approved):
Nice! I like this better than a boolean flag. Allows us to provide additional methods in case we have them in the future.

/lgtm

PR #182, opened 2023-06-14
resmon: fix the error message

ffromani:
It's InfoS, so it uses a different format. Fix accordingly. Note the issue is only cosmetic, the code is NOT crashing.


swatisehgal (approved):
/lgtm

Issue #173, opened 2023-05-01
wrong use of DeferCleanup in the e2e suite


ffromani:
per [docs](https://pkg.go.dev/github.com/onsi/ginkgo/v2#DeferCleanup): 
```
Note that DeferCleanup does not represent a node but rather dynamically generates the appropriate type of cleanup node based on the context in which it is called. As such you must call DeferCleanup within a Setup or Subject node, and not within a Container node. You can learn more about DeferCleanup here: https://onsi.github.io/ginkgo/#cleaning-up-our-cleanup-code-defercleanup
```

We use `DeferCleanup` in `It` blocks, so this is likely wrong. We should review the tests and fix the usage.

Issue #160, opened 2023-02-07
rethink and rewrite the e2e suite


ffromani:
the RTE e2e suite was written long ago, and was shapen by decisions that are no longer relevant or even proven wrong:

- use the k8s e2e framework to be compatible with NFD: no longer needed, eats a lot of CI time to be build, poses constraints on other libraries we can use
- in general the suite was written in a rush, not really following good standards and practices, so we can and should reuse the more modern knowledge

We should rethink and rewrite the suite, to cut down the flakes and to reduce the CI time

PR #194, opened 2023-07-31
e2e: untie k8s e2e framework from test suite

Tal-or:
This PR remove the dependency of the test suite of the Kubernetes e2e test framework.

We replace the framework with local "framework" of our own named `Fixture` which
contains all the different REST clients we need for the tests to run properly.

We were also using the framework log support - now it replaced with `klog`.


Tal-or:
/hold
We need to verify all tests pass.

go.mod
```diff
@@ -10,8 +10,8 @@ require (
 	github.com/k8stopologyawareschedwg/noderesourcetopology-api v0.1.1
 	github.com/k8stopologyawareschedwg/podfingerprint v0.2.2
 	github.com/mdomke/git-semver v1.0.0
-	github.com/onsi/ginkgo/v2 v2.4.0
-	github.com/onsi/gomega v1.23.0
+	github.com/onsi/ginkgo/v2 v2.11.0
```
ffromani, line 13: which 2.11 (or post-2.4 in general) feature do we need?

test/e2e/e2e_suite_test.go
```diff
@@ -17,50 +17,17 @@ limitations under the License.
...
-	}
-
-	rand.Seed(time.Now().UnixNano())
-	os.Exit(m.Run())
-}
-
 func TestE2E(t *testing.T) {
-	e2ekube.RunE2ETests(t)
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "E2e Suite")
```
ffromani, line 32: nit: `E2E` (two capital "e") please

go.mod
```diff
@@ -12,141 +12,79 @@ require (
...
-	k8s.io/api v0.26.4
-	k8s.io/apiextensions-apiserver v0.0.0
-	k8s.io/apimachinery v0.26.4
-	k8s.io/client-go v0.26.4
-	k8s.io/klog/v2 v2.80.1
+	k8s.io/api v0.27.2
+	k8s.io/apiextensions-apiserver v0.27.2
+	k8s.io/apimachinery v0.27.2
+	k8s.io/client-go v0.27.2
+	k8s.io/klog/v2 v2.90.1
```
ffromani, line 24: let's NOT jump to kube 1.27 just yet unless there is a clear and strong reason to, please

test/e2e/utils/fixture/fixture.go
```diff
@@ -0,0 +1,96 @@
...
+	"fmt"
+	topologyclientset "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/generated/clientset/versioned"
+	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"k8s.io/client-go/kubernetes"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
```
ffromani: we need to reorder the imports

test/e2e/utils/fixture/fixture.go
```diff
@@ -0,0 +1,96 @@
...
+
+func init() {
+	// important so we keep the same context across the suite run
+	if initialized {
+		return
+	}
+	F.Ctx = context.Background()
+	Expect(initK8SClient()).ToNot(HaveOccurred())
+	Expect(initClient()).ToNot(HaveOccurred())
+	Expect(initTopologyClient()).ToNot(HaveOccurred())
```
ffromani: IIRC `Expect` outside a `It`(/`Describe`...) block are not legal

test/e2e/utils/fixture/fixture.go
```diff
@@ -0,0 +1,96 @@
...
+var F Fixture
+var initialized bool
+
+type Fixture struct {
+	Ctx     context.Context
+	Cli     client.Client
+	K8SCli  *kubernetes.Clientset
+	TopoCli *topologyclientset.Clientset
+	// TestingNS randomly generated for each test and gets deleted when the test ends
+	TestingNS *corev1.Namespace
```
ffromani, line 29: let's call it `Namespace`, (or just `NS`), the fact is a testing NS is obvious enough from the context

test/e2e/rte/rte.go
```diff
@@ -22,6 +22,7 @@ package rte
 
 import (
 	"context"
+	"github.com/k8stopologyawareschedwg/resource-topology-exporter/test/e2e/utils/fixture"
```
ffromani: need to reorder import

test/e2e/rte/rte.go
```diff
@@ -71,9 +71,9 @@ var _ = ginkgo.Describe("[RTE][InfraConsuming] Resource topology exporter", func
 	ginkgo.BeforeEach(func() {
 		var err error
 
-		err = e2etestns.Setup(f)
+		nsCleanup, err := fixture.F.CreateNamespace("rte")
```
ffromani: I kinda like we seem to want to mimic `testing.T` with `fixture.F`, but I don't like too much relying on a global variable to do this work. It would be nicer if we create our own `fixture.F` on each `BeforeEach` block

test/e2e/topology_updater/topology_updater.go
```diff
@@ -54,18 +51,15 @@ var _ = ginkgo.Describe("[TopologyUpdater][InfraConsuming] Node topology updater
...
 		tmScope             string
-		topologyClient      *topologyclientset.Clientset
 		topologyUpdaterNode *corev1.Node
 		workerNodes         []corev1.Node
 	)
 
-	f := framework.NewDefaultFramework("topology-updater")
-	f.NamespacePodSecurityEnforceLevel = admissionapi.LevelPrivileged
-	f.SkipNamespaceCreation = true
+	f := &fixture.F
```
ffromani, line 58: this is nicer than before, but shouldn't we get a new `fixture.Fixture` (thus a namespace) for each testcase, possibly in `BeforeEach`?

test/e2e/utils/remoteexec/command.go
```diff
@@ -0,0 +1,62 @@
+package remoteexec
```
ffromani, line 1: we can relicense as `kubernetes authors` (I wrote all of it) - let's add boilerplate license text here and for all the code in `test/e2e/utils`

test/e2e/utils/fixture/fixture.go
```diff
@@ -6,6 +6,7 @@ import (
 	topologyclientset "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/generated/clientset/versioned"
 	. "github.com/onsi/gomega"
 	corev1 "k8s.io/api/core/v1"
+	apiextension "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
```
ffromani, line 9: can we squash this commit  with the one which introduces the `Fixture`?

test/e2e/utils/fixture/fixture.go
```diff
@@ -8,6 +8,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apiextension "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/rest"
```
ffromani, line 11: good improvement. Can squash this commit with the one(s) introducing the `fixture`?

ffromani (commented):
initial review, lots of good work here. If we can extract the commits that
1. move from `framework.Logf` to `klog`
2. replace `NotTo` with `ToNot`

we could merge that PR right away

Tal-or:
> initial review, lots of good work here. If we can extract the commits that
> 
> 1. move from `framework.Logf` to `klog`

This is a must in order to get rid of the framework deps. What am I missing here?

test/e2e/topology_updater/topology_updater.go
Tal-or: The `Fixture` object is a global object, initialized once, so whether it would lie in a `BeforeEach` or not it doesn't matter.
The reason I didn't add it in the `BeforeEach` is because I didn't want to imply this object needs re initialization for every run.
The namespace creation is a separate operation that done for every test case by the fact we run it in `BeforeEach`

test/e2e/rte/rte.go
Tal-or: Initializing the `Fixture` for each case seems a bit expensive don't you think?
It essentially mean we need to get the rest config and initialize all the client again and again for each test.
What benefit do we gain here?

BTW the line   nsCleanup, err := fixture.F.CreateNamespace("rte") changed a bit in later commit (mostly cosmetics).

test/e2e/rte/rte.go
Tal-or: done on later PR

go.mod
Tal-or: probably done by mistake when I start using the controller-runtime. will fix

test/e2e/utils/fixture/fixture.go
Tal-or: I saw the k8s e2e framework using similar approach:
https://github.com/openshift-kni/numaresources-operator/blob/ea32cc11d13479c6a4e5091f7c080040bd67d4b1/vendor/k8s.io/kubernetes/test/e2e/framework/log.go#L61 
so  I thought it's OK

test/e2e/rte/rte.go
ffromani: I think we should (strongly) prioritize isolation over speed. Re-initializing clients every time is wasteful indeed and we can probably avoid that using a package to holds clients. We can initialize them explicitely and/or with `init`. This is close to what https://github.com/openshift-kni/numaresources-operator/blob/main/test/utils/clients/clients.go does

Tal-or:
> > > initial review, lots of good work here. If we can extract the commits that
> > > 
> > > 1. move from `framework.Logf` to `klog`
> > 
> > 
> > This is a must in order to get rid of the framework deps. What am I missing here?
> 
> Sorry I was not clear. If we can extract these bits in a new PR (this one will then depend on that), THEN we can merge the new PR immediately and possibly simplify this one. Moving to `klog` and fixing `NotTo` is a good improvement anyway.

Done. #195

ffromani:
Fixes: #173

ffromani:
Fixes: #160

go.mod
Tal-or: The K8s update done automatically when I'm calling `go mod`  I need to figure out why

ffromani: thinking about it could be no biggie if we keep the explicit `replace` pinning deps to 1.26.7

Tal-or: yes scratch it that's what we're doing

ffromani: good, false alarm! Sorry for the noise

test/e2e/rte/metrics.go
```diff
@@ -102,18 +99,9 @@ var _ = ginkgo.Describe("[RTE][Monitoring] metrics", func() {
...
-				PodName:            rtePod.Name,
-				ContainerName:      rteContainerName,
-				Stdin:              nil,
-				CaptureStdout:      true,
-				CaptureStderr:      true,
-				PreserveWhitespace: false,
-			})
-			gomega.Expect(err).ToNot(gomega.HaveOccurred(), "ExecWithOptions failed with %s:\n%s", err, stderr)
+			cmd := []string{"curl", fmt.Sprintf("http://127.0.0.1:%d/metrics", metricsPort)}
+			stdout, stderr, err := remoteexec.CommandOnPod(f.Ctx, f.K8SCli, rtePod, rteContainerName, cmd...)
```
ffromani, line 103: Let's dump the stdout at least initially. Something like
```
stdout, stderr, err := remoteexec.CommandOnPod(f.Ctx, f.K8SCli, rtePod, rteContainerName, "curl", fmt.Sprintf("http://127.0.0.1:%d/metrics", metricsPort))
klog.Infof("curl stdout BEGIN:\n%s\ncurl stdout END", stdout)
klog.Infof("curl stderr BEGIN:\n%s\ncurl stderr END", stderr)
```

test/e2e/rte/metrics.go
```diff
@@ -125,29 +113,23 @@ var _ = ginkgo.Describe("[RTE][Monitoring] metrics", func() {
...
```
ffromani, line 131: same as above

test/e2e/rte/metrics.go
Tal-or: I actually thought about that but since it's done in the `Eventually` wouldn't it spam the logs badly?

Tal-or: I dump the latest `stderr`\`stdout` in case of failure

ffromani:
the metrics tests is flaking, but it was flaking also before. Shaving almost 3m from e2e execution time is an achievement and this alone is worth the price. Solid work, let's merge it and iterate.

ffromani:
before: 
image

after:
image

PR #195, opened 2023-07-31
e2e: perquisites for untiying k8s e2e framework 

Tal-or:
General improvements and logger replacement for setting up the stage to get rid of k8s e2e framework


ffromani (approved):
nice work!

Issue #175, opened 2023-05-03
update base image


ffromani:
we should use something more modern than centos7.9

PR #196, opened 2023-08-01
images: bump to centos8

ffromani:
The centos 7 image is too old, let's bump to 8. We will probably update again soon, but one step at time.

Fixes: #175


swatisehgal (approved):
/lgtm

PR #198, opened 2023-08-02
e2e:fixture: initialize fixture explicitly

Tal-or:
Initialize the fixture explicitly and not automatically just by importing it.a

This is because of other projects which are importing the suite and not necessarily using it in the same environment, which in turn might cause the fixture initialization to failed.


test/e2e/utils/fixture/fixture.go
```diff
@@ -17,7 +17,7 @@ import (
 	topologyclientset "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/generated/clientset/versioned"
 )
 
-var F Fixture
+var f Fixture
```
ffromani, line 20: as much as I love to use name as terse as possible, can we rename this `fxt` to make it more grep-friendly?

also see: https://google.github.io/styleguide/go/decisions#variable-names

ffromani (commented):
LGTM with just one comment inside

ffromani:
we will address the metrics test flakiness once we merge all the fixes to make integration of the testsuite easier

test/e2e/utils/fixture/fixture.go
Tal-or: Sure. thank you always for the added value!

ffromani (approved):
thanks for addressing the feedback so quickly!

PR #201, opened 2023-08-03
e2e: utils: nodes: improve RTE metrics test troubleshooting

ffromani:
If memory serves, metrics just always take some time before to be populated (that's expected, documented and bening, but of course I don't have docs handy). If the metrics test executes too early, because ginkgo random ordering, that could explain the flakiness.

This PR won't fix the flakiness but will make it easier to understand if this is indeed the case.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -83,7 +83,7 @@ jobs:
     - name: run E2E tests
       run: |
         export KUBECONFIG=${HOME}/.kube/config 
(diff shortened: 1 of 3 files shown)


Tal-or:
/lgtm
/approve

Now we need to trigger the tests for couple of times and see if we rule out the flakiness

ffromani:
retried 3 times in a row - at least is not worse than before (https://github.com/k8stopologyawareschedwg/resource-topology-exporter/actions/runs/5747761189)

PR #206, opened 2023-08-09
podres: client: make independent of k8s code 

ffromani:
It's actively discouraged by k8s project to import `k8s.io/kubernetes`
packages, and, in addition to this, it's recommended (https://github.com/kubernetes/kubernetes/blob/v1.28.0-rc.0/pkg/kubelet/apis/podresources/client.go#L32)
to reimplement the client in each consuming project, copying and pasting the related functions.

We missed this recommentation up until now, so we do it now.
This way, we can remove the latest dep to `k8s.io/kubernetes`,
have less deps overall and a leaner `vendor/`.


swatisehgal (approved):
/lgtm
/approve

PR #207, opened 2023-08-10
kubernetes: bundle, not vendor, cpuset code

ffromani:
in order to remove all dependencies on `k8s.io/kubernetes`, bundle (= copy/paste in the tree) the cpuset code instead of vendoring. We can just revert back to vendoring when we can update to `k8s.io/utils` compatible to 1.28, so worst case when we bump to kube 1.28.


ffromani:
final step and the engine that is propelling all this work: #206 (depends on this one, WIP)

Tal-or:
What is the justification of getting rid of the k8s deps?

ffromani:
> What is the justification of getting rid of the k8s deps?

The main driver is to be able to get rid of `import k8s.io/kubernetes`, which we did in the past and which we do in other project. Importing kube packages as libraries this way is not supported (actually discouraged) from k8s project and can break anytime. Our saving grace was (is) we just need a very narrow set of packages which happen to be quite stable.

Tal-or:
> > What is the justification of getting rid of the k8s deps?
> 
> The main driver is to be able to get rid of `import k8s.io/kubernetes`, which we did in the past and which we do in other project. Importing kube packages as libraries this way is not supported (actually discouraged) from k8s project and can break anytime. Our saving grace was (is) we just need a very narrow set of packages which happen to be quite stable.

So why not just wait for 1.28 and it will remove automatically because `cpuset` goes under `k8s/utils`?

ffromani:
> > > What is the justification of getting rid of the k8s deps?
> > 
> > 
> > The main driver is to be able to get rid of `import k8s.io/kubernetes`, which we did in the past and which we do in other project. Importing kube packages as libraries this way is not supported (actually discouraged) from k8s project and can break anytime. Our saving grace was (is) we just need a very narrow set of packages which happen to be quite stable.
> 
> So why not just wait for 1.28 and it will remove automatically because `cpuset` goes under `k8s/utils`? we'll still need to change the imports path, but we'll do that anyway according to the PR description.

that's also an option. It's up to us (maintaining the project) to decide if we want to have this bundling (expected to be stable and low-effort) for a while or just wait for 1.28.
Just note this PR is a prerequisite for https://github.com/k8stopologyawareschedwg/resource-topology-exporter/pull/206

Tal-or:
/lgtm
/approve
Then let's move on and merge it. 
When 1.28 time comes we'll deal it

PR #211, opened 2023-08-21
notification: statedirs: drop feature

ffromani:
Remove support to get notifications watching the kubelet state dirs. As far as kubelet is implemented, it's hard to secure, and has issues anyway (cannot detect deletes).
A better replacement is use a OCI hook to leverage the notification file.

This feature has been deprecated since 0.10.z, time to remove it for good.


swatisehgal (approved):
/lgtm

Issue #203, opened 2023-08-04
Set OwnerReference for NRT objects


ffromani:
Should we follow the example of the [DRA example driver](https://github.com/kubernetes-sigs/dra-example-driver) and set OwnerReferences backlinking to nodes to the NRT objects?

jlojosnegros:
I gonna take this, if that is ok.

Is the target to add the Node as Owner of the NRT to trigger deletion if a node is deleted from the cluster or the discussion about if doing so its ok?

ffromani:
go for it! the main benefit of this change is to avoid leftovers in the cluster: if a node object gets deleted, the corresponding NRT should be deleted as well (no stale objects).
Truth to be told, I expect this flow to trigger extremely rarely, but still why not?

jlojosnegros:
/close 
as #212  is merged

PR #212, opened 2023-09-06
topology-updater: Add Node OwnerReference to NRTs

jlojosnegros:
Add the corresponding Node as OwnerReference in thr NRT, so it gets deleted if the Node object gets deleted, to avoid leftovers in the cluster.

address: #203


pkg/nrtupdater/nrtupdater.go
```diff
@@ -61,11 +64,12 @@ func (mi MonitorInfo) UpdateReason() string {
 	return RTEUpdateReactive
 }
 
-func NewNRTUpdater(args Args, tmconf TMConfig) *NRTUpdater {
+func NewNRTUpdater(k8sInterface kubernetes.Interface, args Args, tmconf TMConfig) *NRTUpdater {
```
ffromani, line 67: I think is better if the `NRTUpdater` takes and uses a more limited interface like
```
type NodeGetter interface {
	Get(ctx context.Context, name string, opts metav1.GetOptions) (*v1.Node, error)
}
```
This is not just to fulfill the principle of minium privilege, but also because I'm reluctant to require this functionality a Get to a node for each update. I'm even wary about adding a watch.
So I'd start decoupling the responsabilities. The NRTUpdater just needs a way to access the node metadata (uid...) from name. Another component should be responsible to pull this data and keep it uptodate (or not)

pkg/nrtupdater/nrtupdater.go
```diff
@@ -130,6 +134,63 @@ func (te *NRTUpdater) updateNRTInfo(nrt *v1alpha2.NodeResourceTopology, info Mon
...
+	}
+
+	// look for all the OwnerReferences pointing to a Node
+	nodeOwnerReferences := []metav1.OwnerReference{}
+	for _, own := range nrt.OwnerReferences {
+		if own.Kind == "Node" {
+			nodeOwnerReferences = append(nodeOwnerReferences, own)
+		}
+	}
+	numnodeowners := len(nodeOwnerReferences)
```
ffromani, line 165: RTE owns the NRT objects so I think you can just force the OwnerReference field the way you want. No need to be so nice (also applies to the lines below)

pkg/resourcetopologyexporter/resourcetopologyexporter.go
```diff
@@ -61,7 +61,7 @@ func Execute(hnd resourcemonitor.Handle, nrtupdaterArgs nrtupdater.Args, resourc
 	}
 	go resObs.Run(eventSource.Events(), condChan)
 
-	upd := nrtupdater.NewNRTUpdater(nrtupdaterArgs, tmConf.config)
+	upd := nrtupdater.NewNRTUpdater(hnd.K8SCli, nrtupdaterArgs, tmConf.config)
```
ffromani, line 64: this should be done conditionally controlled by an option. I'm very wary of adding load to the apiserver

pkg/nrtupdater/nrtupdater.go
jlojosnegros: Makes sense. 
I have tried to implement it in a newly added commit

pkg/nrtupdater/nrtupdater.go
jlojosnegros: Ok,  updated in the new commit also.

pkg/resourcetopologyexporter/resourcetopologyexporter.go
jlojosnegros: Ok, updated in the new commit also.

pkg/config/config.go
```diff
@@ -111,6 +111,7 @@ func LoadArgs(args ...string) (ProgArgs, error) {
 	flags.StringVar(&pArgs.RTE.KubeletConfigFile, "kubelet-config-file", "/podresources/config.yaml", "Kubelet config file path.")
 	flags.StringVar(&pArgs.RTE.PodResourcesSocketPath, "podresources-socket", "unix:///podresources/kubelet.sock", "Pod Resource Socket path to use.")
 	flags.BoolVar(&pArgs.RTE.PodReadinessEnable, "podreadiness", true, "Custom condition injection using Podreadiness.")
+	flags.BoolVar(&pArgs.RTE.OwnerReferenceEnable, "node-owns-nrt", false, "RTE will inject NRT's related node as OwnerReference to ensure cleanup if the node is deleted.")
```
ffromani, line 114: I'd rename the option like `add-nrt-owner`

pkg/nrtupdater/node_getter.go
```diff
@@ -0,0 +1,68 @@
...
+	return fmt.Sprintf("error connection k8s: %v", err.Err)
+}
+func (err ConnectionError) Unwrap() error {
+	return err.Err
+}
+
+func IsNotConfiguredNodeGetter(err error) bool {
+	_, ok := err.(NotConfigured)
+	return ok
+}
```
ffromani, line 32: we probably want to use the more modern idiom `errors.Is` https://go.dev/blog/go1.13-errors

pkg/nrtupdater/node_getter.go
```diff
@@ -0,0 +1,68 @@
...
+	ng.node = node.DeepCopy()
+	return ng.node, nil
+}
+
+func NewNodeGetter(enabled bool, k8xcli kubernetes.Interface) NodeGetter {
+	if enabled {
+		return &OneshotNodeGetter{K8SCli: k8xcli}
+	}
+	return &DisabledNodeGetter{}
+}
```
ffromani, line 68: thanks, this is the right direction. Even better would be to have an object which wraps a map nodeName -> nodeObject, initialize it with the output of `List` (once at startup) and return the cached data with `Get`.
Nodes should come and go rarely, and if updates are needed we can just restart the RTE.
Once this is in place, a followup could be a GC process (compare the node listing with the NRT listing) like NFD implemented lately. Stuff for a followup PR.

ffromani (commented):
thanks, we're heading the right direction. Few more comments; next up, I'd like a e2e test added.

pkg/nrtupdater/node_getter.go
jlojosnegros: Thanks for the link

pkg/nrtupdater/node_getter.go
```diff
@@ -34,27 +42,34 @@ func (ng *DisabledNodeGetter) Get(ctx context.Context, nodeName string, opts met
...
+	nodes map[string]*corev1.Node
 }
 
-func (ng *OneshotNodeGetter) Get(ctx context.Context, nodeName string, opts metav1.GetOptions) (*corev1.Node, error) {
-	if ng.node != nil {
-		return ng.node, nil
+func NewCachedNodeGetter(k8sInterface kubernetes.Interface, ctx context.Context) *CachedNodeGetter {
+	nodelist, err := k8sInterface.CoreV1().Nodes().List(ctx, metav1.ListOptions{})
+	if err != nil {
+		return nil
```
jlojosnegros, line 52: Not completely ok with returning nil here.

I was thinking about checking in the `Get` if the map has been populated and retry again if not. 
Problem is that option would mean to check a boolean everytime, even if the map already has data, and make the first call take more time. 

In the end I kept `Get` as simple as possible and let client to handle the nil 
I have defaulted to `disable` in the resourcetopologyexporter.go

pkg/config/config.go
jlojosnegros: done

pkg/nrtupdater/node_getter.go
```diff
@@ -39,7 +31,7 @@ type DisabledNodeGetter struct {
 }
 
 func (ng *DisabledNodeGetter) Get(ctx context.Context, nodeName string, opts metav1.GetOptions) (*corev1.Node, error) {
-	return nil, NotConfigured{}
+	return nil, fmt.Errorf("%w", NotConfigured)
```
ffromani, line 42: it seems to me you can just return `NotConfigured` here

pkg/nrtupdater/node_getter.go
```diff
@@ -34,27 +42,34 @@ func (ng *DisabledNodeGetter) Get(ctx context.Context, nodeName string, opts met
...
-		return ng.node, nil
+func NewCachedNodeGetter(k8sInterface kubernetes.Interface, ctx context.Context) *CachedNodeGetter {
+	nodelist, err := k8sInterface.CoreV1().Nodes().List(ctx, metav1.ListOptions{})
+	if err != nil {
+		return nil
+	}
+
+	retVal := &CachedNodeGetter{nodes: make(map[string]*corev1.Node, len(nodelist.Items))}
+	for _, node := range nodelist.Items {
+		retVal.nodes[node.Name] = &node
```
ffromani, line 57: this is dangerous and could hit a common go gotcha: https://github.com/golang/go/wiki/CommonMistakes

the fix is simple:
```
	for idx := range nodelist.Items {
                node := &nodelist.Items[idx]
		retVal.nodes[node.Name] = node
```

pkg/nrtupdater/node_getter.go
```diff
@@ -34,27 +42,34 @@ func (ng *DisabledNodeGetter) Get(ctx context.Context, nodeName string, opts met
...
+	return retVal
+}
+
+func (ng *CachedNodeGetter) Get(ctx context.Context, nodeName string, _ metav1.GetOptions) (*corev1.Node, error) {
+	if node, found := ng.nodes[nodeName]; found {
+		return node, nil
 	}
-	ng.node = node.DeepCopy()
-	return ng.node, nil
+	return nil, fmt.Errorf("%w", NotFound{NodeName: nodeName})
```
ffromani, line 67: I don't think there's need to wrap, I think you can just return `NotFound{NodeName: nodeName}`

pkg/nrtupdater/nrtupdater_test.go
```diff
@@ -51,8 +52,8 @@ func TestUpdateTMPolicy(t *testing.T) {
 
 	var err error
 	k8sClient := clientk8sfake.NewSimpleClientset()
-	nodeGetter := OneshotNodeGetter{K8SCli: k8sClient}
-	nrtUpd = NewNRTUpdater(&nodeGetter, args, tmConfInitial)
+	nodeGetter := NewCachedNodeGetter(k8sClient, context.TODO())
```
ffromani, line 55: `context.Background()` has the same behavior but conveys better semantic: we're fine with unbounded wait (no cancelation), and this is what we want (vs TODO which is meant to be a temporary placeholder, but oftentimes abused as synonym for Background)

pkg/nrtupdater/nrtupdater_test.go
```diff
@@ -177,8 +178,8 @@ func TestUpdateOwnerReferences(t *testing.T) {
 	cli := fake.NewSimpleClientset()
 	var err error
 	k8sClient := clientk8sfake.NewSimpleClientset(&node)
-	nodeGetter := OneshotNodeGetter{K8SCli: k8sClient}
-	nrtUpd = NewNRTUpdater(&nodeGetter, args, tmConfig)
+	nodeGetter := NewCachedNodeGetter(k8sClient, context.TODO())
```
ffromani, line 181: same

pkg/resourcetopologyexporter/resourcetopologyexporter.go
```diff
@@ -62,7 +63,11 @@ func Execute(hnd resourcemonitor.Handle, nrtupdaterArgs nrtupdater.Args, resourc
 	}
 	go resObs.Run(eventSource.Events(), condChan)
 
-	nodeGetter := nrtupdater.NewNodeGetter(rteArgs.AddNRTOwnerEnable, hnd.K8SCli)
+	nodeGetter := nrtupdater.NewNodeGetter(rteArgs.AddNRTOwnerEnable, hnd.K8SCli, context.TODO())
+	if nil == nodeGetter {
+		klog.V(2).Info("Cannot enable 'add-nrt-owner'.")
```
ffromani, line 68: I'd just error out here. We may move this block up to not leak `go resObs.Run()` or anything else

ffromani (commented):
thanks, getting there.

pkg/nrtupdater/node_getter.go
jlojosnegros: Yes, I did it that way following one of the sentences in the article about errors
"""
If a function is defined as returning an error wrapping some sentinel or type, do not return the underlying error directly.
```
var ErrPermission = errors.New("permission denied")

// DoSomething returns an error wrapping ErrPermission if the user
// does not have permission to do something.
func DoSomething() error {
    if !userHasPermission() {
        // If we return ErrPermission directly, callers might come
        // to depend on the exact error value, writing code like this:
        //
        //     if err := pkg.DoSomething(); err == pkg.ErrPermission { … }
        //
        // This will cause problems if we want to add additional
        // context to the error in the future. To avoid this, we
        // return an error wrapping the sentinel so that users must
        // always unwrap it:
        //
        //     if err := pkg.DoSomething(); errors.Is(err, pkg.ErrPermission) { ... }
        return fmt.Errorf("%w", ErrPermission)
    }
    // ...
}
```
"""

ffromani: ok, good find. TIL.

pkg/resourcetopologyexporter/resourcetopologyexporter.go
jlojosnegros: ok ... I didn't do it because I was not sure if a fail on that small feature was worthy to fail .

pkg/nrtupdater/node_getter.go
ffromani: explained in https://github.com/k8stopologyawareschedwg/resource-topology-exporter/pull/212#discussion_r1319842202 . It's more correct this way.

tools/nrtstress/main.go
```diff
@@ -53,6 +55,8 @@ func main() {
 		Policy: tmPolicy,
 		Scope:  tmScope,
 	}
-	upd := nrtupdater.NewNRTUpdater(nrtupdaterArgs, tmConf)
+
+	nodeGetter := nrtupdater.NewCachedNodeGetter(clientk8sfake.NewSimpleClientset(), context.Background())
```
ffromani, line 59: I'd use the NOP/null/void getter in `nrtstress`

pkg/resourcetopologyexporter/resourcetopologyexporter.go
```diff
@@ -40,6 +42,12 @@ func Execute(hnd resourcemonitor.Handle, nrtupdaterArgs nrtupdater.Args, resourc
 		return err
 	}
 
+	nodeGetter := nrtupdater.NewNodeGetter(rteArgs.AddNRTOwnerEnable, hnd.K8SCli, context.TODO())
+	if nil == nodeGetter {
```
ffromani, line 46: we don't usually set yoda conditions

pkg/nrtupdater/node_getter.go
jlojosnegros: Thanks for the catch and the link.

test/e2e/topology_updater/topology_updater.go
```diff
@@ -96,6 +97,33 @@ var _ = ginkgo.Describe("[TopologyUpdater][InfraConsuming] Node topology updater
...
+					klog.Infof("Too many OwnerReferences: expected=1; got=%d", len(finalNodeTopo.OwnerReferences))
+					klog.Infof("%+#v", finalNodeTopo.OwnerReferences)
+					return false
+				}
+				if !reflect.DeepEqual(finalNodeTopo.OwnerReferences[0], expectedOwnerReference) {
+					klog.Infof("Unexpected OwnerReference: expected=%+#v; got=%+#v", expectedOwnerReference, finalNodeTopo.OwnerReferences[0])
+					return false
+				}
+				return true
+			}, 5*timeout, 5*time.Second).Should(gomega.BeTrue(), "didn't get updated node topology info")
```
ffromani, line 125: please let's use `WithPolling` and `WithTimeout` in new code like in https://github.com/k8stopologyawareschedwg/resource-topology-exporter/blob/main/test/e2e/rte/rte.go#L152
(yes, we should send a followup PR later on to modernize the usages)

ffromani: https://github.com/k8stopologyawareschedwg/resource-topology-exporter/pull/214

jlojosnegros: done, Thanks for the comment.

pkg/nrtupdater/node_getter.go
```diff
@@ -0,0 +1,80 @@
...
+func NewNodeGetter(enabled bool, k8xcli kubernetes.Interface, ctx context.Context) NodeGetter {
+	if enabled {
+		ret := NewCachedNodeGetter(k8xcli, ctx)
+		if ret == nil {
+			return nil
+		}
+		return ret
+	}
+	return &DisabledNodeGetter{}
+}
```
ffromani, line 80: I'm not convinced about this wrapper. It seems to be used once, and I'm not sold about swallowing the error. I think it would be better to inline in the calling site, possibly bubbling up the error.

ffromani (commented):
thanks, all the pieces are now in there. Just a couple of last questions and we can merge

ffromani (approved):
thanks for all the updates!

jlojosnegros:
> thanks for all the updates!

Thank you for all you comments and reviews! :)

PR #213, opened 2023-09-06
toolchain: update toolchain to 1.20

jlojosnegros:
Update go toolchain to 1.20.7

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -29,7 +29,7 @@ jobs:
       uses: actions/setup-go@v3
       id: go
       with:
-        go-version: 1.19.3
+        go-version: 1.20.7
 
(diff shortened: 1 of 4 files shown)


ffromani:
thanks for this update! I see you are both updating the toolchain and updating deps, what's the benefit of this? can we split in 1. updating the toolchain and 2. updating the deps?

jlojosnegros:
> thanks for this update! I see you are both updating the toolchain and updating deps, what's the benefit of this? can we split in 1. updating the toolchain and 2. updating the deps?

You're right, better to update deps in another PR if needed at all.

ffromani:
LGTM, but let's cut 0.14.0 before

ffromani (approved):
LGTM

PR #222, opened 2023-09-22
golangci-lint add developer config and related fixes

ffromani:
Add new makefile target to run golangci-lint and to be used by developers.
Later (soonish) we will integrate golangci-lint in CI using the recommended GH action.

This PR also includes a batch of related fixes for the issues raised by the golangci-lint run.


swatisehgal (approved):
/lgtm
/approve

PR #239, opened 2023-11-09
[release-0.11] remove k8s deps

ffromani:
backport of all the changes which allow us to drop k8s deps.
This is useful for users which won't or can't update past 1.25.z.


swatisehgal (approved):
/lgtm

PR #244, opened 2023-12-11
sysinfo: hugepages: stricter checks, louder error handling

ffromani:
Add more validations when checking for hugepages, reducing the chance to read random junk in the system.
The layout of sysfs is stable and guaranteed; thus, should we encounter errors navigating it, we should
make sure to make noise and bubble them up, because it means something really weird happened.

### DIFF
--- a/pkg/sysinfo/hugepages.go
+++ b/pkg/sysinfo/hugepages.go
@@ -21,6 +21,8 @@ import (
 	"strconv"
 	"strings"
 
+	"k8s.io/klog/v2"
+
 	"k8s.io/apimachinery/pkg/api/resource"
 )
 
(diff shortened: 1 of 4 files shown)


pkg/sysinfo/hugepages.go
```diff
@@ -46,13 +46,13 @@ func GetHugepages(hnd Handle) ([]Hugepages, error) {
 		if entry.IsDir() && strings.HasPrefix(entryName, "node") {
 			nodeID, err := strconv.Atoi(entryName[4:])
 			if err != nil {
-				klog.Warningf("cannot detect the node ID for %q", entryName)
-				continue
+				klog.Errorf("cannot detect the node ID for %q", entryName)
```
Tal-or, line 49: Is this change going to break the existing functionality?

pkg/sysinfo/hugepages.go
```diff
@@ -72,6 +72,10 @@ func HugepagesForNode(hnd Handle, nodeID int) ([]Hugepages, error) {
 		return nil, err
 	}
 	for _, entry := range entries {
+		if !entry.IsDir() {
```
Tal-or, line 75: Is this deserve log? genuinely asking

ffromani: actually yes, but not erroring out. Adding.

pkg/sysinfo/hugepages.go
ffromani: It should not, because the structure should be stable (and the code should handle it). If it does, better than silently swallowing.

swatisehgal (approved):
/lgtm

PR #250, opened 2024-01-22
WIP: config: configfiles support

ffromani:
Add support for configfiles + configlet (conf.d/*)

WIP TBD


pkg/config/cfgdispatch.go
```diff
@@ -0,0 +1,188 @@
...
+		return 0, nil
+	}
+	j, ok := val.(float64)
+	if !ok {
+		return 0, fmt.Errorf("key %q has non-float64 value representation %T", key, val)
+	}
+	return time.Duration(j), nil
+}
+
+func dispatchConfObj(obj map[string]interface{}, pArgs *ProgArgs) error {
```
Tal-or, line 76: IIUC, the object we expect here is bunch of `yaml` files serialized into dotted format composed into `ProgArgs` object. 
Why can't we just unmarshal the yaml into `ProgArgs`?

pkg/config/cfgfile.go
```diff
@@ -42,11 +44,81 @@ func FixExtraConfigPath(configRoot string) string {
...
+	obj := make(map[string]interface{})
+	data, err := os.ReadFile(configPath)
+	if err != nil {
+		return err
+	}
+	err = yaml.Unmarshal(data, &obj)
+	if err != nil {
+		return err
+	}
+	flat, err := flatten.Flatten(obj, "", flatten.DotStyle)
```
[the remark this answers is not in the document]
ffromani: good point. Added comments to provide rationale.

pkg/config/cfgdispatch.go
ffromani: explained in the newly-added comments

PR #271, opened 2024-03-05
[release-0.18][manual] align with fixes from main

ffromani:
backport fixes from main - cumulative cherry-picks PR


ffromani:
yuck, wrong branch. Redoing. Main will be a bit messy but no biggie

ffromani:
Even though this is annoying, it's harmless and mostly noise in the git history. I prefer NOT to force-push and keep the little mess as reminder to be more careful in the future (and perhaps this is the last push to automate the backports)

PR #273, opened 2024-04-03
Ensure that RTE metrics are served securely with kube-rbac-proxy

swatisehgal:
Add Kube-rbac-proxy side car container to RTE daemonset to ensure that RTE metrics are served securely.


swatisehgal:
Hmm, this is strange.
The cluster ready step (/check-ds.sh) seems to succeed in case of  e2e-ic job but fails for e2e-metrics-https job!

swatisehgal:
/close

Kube-rbac-proxy was considered a suitable way to expose metrics over https for the ease of backportability.
For now, we already have metrics served over https  in 0.19,0.18 and 0.17 versions and if this support is a requirement we can expect the consumers to upgrade to one of these three version.

PR #274, opened 2024-04-03
Increase timeout to allow enough time for ds pod creation

swatisehgal:
There can be scenarios where Daemon set has been successfully created but the pods haven't.

We increase the retry attempts to give more time for the corresponding pods under the RTE daemonset to be created.

Rather than waiting for 2 minutes (24 attempts every 5 secs), we wait for 5 minutes (60 attempts every 5 secs).

### DIFF
--- a/hack/check-ds.sh
+++ b/hack/check-ds.sh
@@ -6,7 +6,7 @@ DAEMONSET="${2:-resource-topology-exporter-ds}"
 FAILED=1
 
 function wait-for-daemonset(){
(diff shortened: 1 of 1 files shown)


ffromani (approved):

ffromani:
CI is harsh so I'm fine with extending the timeout, but with 5m timeout we exhausted the goodwill bonus. If still flakes we will need to deep dive future failures.

PR #280, opened 2024-04-08
Add test cases for explicit Metric Address to be executed in CI

swatisehgal:
Since it is assumed that the RTE pod is created prior to execution of the e2e tests, we add additional CI lanes to tackle cases where Metric Address and port is explicitly specified. We add two cases, one where the metric mode is `http` and another one where `httptls`.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -98,7 +98,11 @@ jobs:
       run: |
         kubectl logs -l name=resource-topology -c resource-topology-exporter-container || :
 
-  e2e-metrics-https:
+  e2e-metrics:
+    strategy:
+      matrix:
+        mode: [http, httptls]
+        address: ["0.0.0.0","127.120.110.100"]
     runs-on: ubuntu-22.04
     env:
       E2E_NODE_REFERENCE: true
@@ -106,7 +110,9 @@ jobs:
       E2E_TOPOLOGY_MANAGER_SCOPE: container
       RTE_CONTAINER_IMAGE: quay.io/k8stopologyawarewg/resource-topology-exporter:ci
       RTE_METRICS_CLI_AUTH: false
-      RTE_METRICS_MODE: httptls
+      RTE_METRICS_MODE: ${{ matrix.mode }}
+      METRICS_ADDRESS: ${{ matrix.address }}
+      METRICS_PORT: "2112"
       RTE_POLL_INTERVAL: 10s
       RTE_VERBOSE: 6
     steps:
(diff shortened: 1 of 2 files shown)


swatisehgal:
@ffromani PTAL and let me know if this aligns with what you were expecting with the e2e tests. The monitoring e2e already has all the code we need so all we needed was to update the parameters and specify environment variables corresponding to the Address and port.

ffromani (commented):
I wasn't expecting to fix using just the github config, but this fully makes sense so I'm fine with it. If you're up to, we can use `matrix` and simplify a bit our config: https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs

the matrix variables can be `RTE_METRICS_MODE` and `METRICS_ADDRESS`
(we should rename `METRICS_ADDRESS` and `METRICS_PORT` in the future for consistency).
about changing the IP, we can probably get away with something like `127.120.110.100` since localhost has a /8 subnet

swatisehgal:
> I wasn't expecting to fix using just the github config, but this fully makes sense so I'm fine with it.

Yep, I assumed that I would need to add a new e2e test but after reviewing the code base this seemed like the right way to go.

> If you're up to, we can use `matrix` and simplify a bit our config: https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs
> 
> the matrix variables can be `RTE_METRICS_MODE` and `METRICS_ADDRESS` (we should rename `METRICS_ADDRESS` and `METRICS_PORT` in the future for consistency). 

TIL, sure will play around with matrix!

> about changing the IP, we can probably get away with something like `127.120.110.100` since localhost has a /8 subnet

Thanks!

.github/workflows/e2e.yml
```diff
@@ -98,15 +98,19 @@ jobs:
       run: |
         kubectl logs -l name=resource-topology -c resource-topology-exporter-container || :
 
-  e2e-metrics-https:
+  example_matrix:
```
ffromani, line 101: `e2e-metrics` perhaps? Or anything you like without "example" :)

swatisehgal: Yep, fixed it. Thanks!

.github/workflows/e2e.yml
```diff
@@ -176,3 +176,169 @@ jobs:
...
+  e2e-metrics-explicit-address-https:
+    runs-on: ubuntu-22.04
+    env:
+      E2E_NODE_REFERENCE: true
+      E2E_TOPOLOGY_MANAGER_POLICY: single-numa-node
+      E2E_TOPOLOGY_MANAGER_SCOPE: container
+      RTE_CONTAINER_IMAGE: quay.io/k8stopologyawarewg/resource-topology-exporter:ci
+      RTE_METRICS_CLI_AUTH: false
+      RTE_METRICS_MODE: httptls
+      METRICS_ADDRESS: "192.168.0.2"
```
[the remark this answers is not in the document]
swatisehgal: Same as above.

ffromani (commented):
LGTM

very nice!

PR #283, opened 2024-04-10
[release-0.17][manual] ghactions: bump kind image

ffromani:
use image recommended with kind 0.22


ffromani:
red herring, we found the root issue and fixed see #285 and similar PRs

PR #288, opened 2024-04-10
ghactions: more rte logs on error

ffromani:
turns out that if we target a specific pod we get the full available logs, vs a reduced log amount if we target by label

since in CI the RTE pod name is fairly predicatable and we have few pods, let's find (hackishly) the pod name and let's target it to get more context.


swatisehgal (approved):
/lgtm

Good to know about this! Thanks for the PR.

Issue #293, opened 2024-05-21
Get rid of vendoring


jaypipes:
We (Microsoft Network Cloud) have a mirror of this repo internally for software supply chain purposes and the `vendor/` directory makes the mirror maintenance unnecessarily painful. Modern Go and modules are a cleaner, more maintainable way to manage dependencies.

ffromani:
I do have nothing against removing `vendor/` in principle. Let's see how we want to do it.

ffromani:
closed thanks to #294

PR #294, opened 2024-05-21
remove all vendoring

jaypipes:
The `vendor/` directory makes mirror maintenance unnecessarily painful. Modern Go and modules are a cleaner, more maintainable way to manage dependencies.

This patch removes all use of Go vendoring and updates the Makefile targets to install dependent test tooling like `mockery` and `git-semver` into the `_out/` directory instead of looking in a `vendor/` directory.

Issue #293


Makefile
```diff
@@ -140,13 +138,15 @@ _out/golangci-lint: outdir
 		echo "Using golangci-lint cached at $(GOLANGCI_LINT_BIN)";\
 	fi
 
-
 _out/git-semver: outdir
-	@go build -o _out/git-semver vendor/github.com/mdomke/git-semver/main.go
+	@GOBIN=$(shell pwd)/_out go install github.com/mdomke/git-semver/v6@latest
```
ffromani, line 142: floating tags created a lot of pain in the past. Let's use `v6@v6.9.0` or any fixed, non-floating tag you prefer.

ffromani (commented):
LGTM once we address the floating tag comment

Makefile
jaypipes: fixed! :)

ffromani (approved):
thanks, LGTM

PR #303, opened 2024-11-26
security: validate root config path

Tal-or:
The ConfigRoot path is configurable by the user.
This may result in a Path Traversal vulnerability and allow an attacker to read arbitrary files.

We need to validate the path provided by the user is under the allowed list of directories.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -20,6 +20,8 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"path/filepath"
+	"strings"
 
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/yaml"
@@ -107,6 +109,10 @@ func LoadArgs(args ...string) (ProgArgs, error) {
 		return pArgs, err
 	}
 
+	configRoot, err = ValidateConfigRootPath(configRoot)
+	if err != nil {
+		return pArgs, err
+	}
 	// now the real processing begins. From now on we waste nothing
 	if pArgs.Global.Debug {
 		klog.Infof("configRoot=%q extraConfigPath=%q", configRoot, extraConfigPath)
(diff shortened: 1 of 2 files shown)


Tal-or:
I started with `/etc, /tmp` as allowed directory to point the config files to. 
We may add more options if needed.

Tal-or:
/hold
Need to validate it solves the security vulnerability

pkg/config/config.go
```diff
@@ -150,3 +155,22 @@ func Finalize(pArgs *ProgArgs) error {
 	}
 	return err
 }
+
+func ValidateConfigRootPath(configRoot string) error {
+	allowedPrefixes := []string{
+		"/etc",
+		"/tmp",
+		"/home",
+	}
```
ffromani, line 164: let's start with these prefixes:
`/etc/rte`, `/run/rte`, `/var/rte`, `/usr/local/etc/rte`

pkg/config/config.go
```diff
@@ -150,3 +156,27 @@ func Finalize(pArgs *ProgArgs) error {
...
 	return err
 }
+
+func ValidateConfigRootPath(configRoot string) (string, error) {
+	allowedPrefixes := []string{
+		"/etc/rte",
+		"/run/rte",
+		"/var/rte",
+		"/usr/local/etc/rte",
+		"/home", // mostly for tests
```
ffromani, line 166: ok, but then should be something like `/home/*/.rte`

pkg/config/config.go
```diff
@@ -150,3 +156,27 @@ func Finalize(pArgs *ProgArgs) error {
...
+	}
+	if configRoot == "" {
+		return configRoot, fmt.Errorf("configRoot is not allowed to be an empty string")
```
ffromani, line 169: return explicit `""` here (mostly to make the tool happy)

pkg/config/config.go
```diff
@@ -150,3 +156,27 @@ func Finalize(pArgs *ProgArgs) error {
...
+	}
+	// Resolve and clean the input path
+	absPath, err := filepath.Abs(filepath.Clean(configRoot))
+	if err != nil {
+		return configRoot, fmt.Errorf("failed to validate configRoot path: %w", err)
```
ffromani, line 174: return `""` here (mostly to [try to] make the tool happy)

pkg/config/config.go
```diff
@@ -150,3 +156,27 @@ func Finalize(pArgs *ProgArgs) error {
...
+	}
+	for _, prefix := range allowedPrefixes {
+		if strings.HasPrefix(absPath, prefix) {
+			return absPath, nil
```
ffromani, line 178: if the prefix is matching, let's not return the original path (even though is correct and we know it) but let's build a path and return it re-creating the user-supplied path (!!).
This should make the tool happy. I hope.

pkg/config/config.go
Tal-or: This is going to break the existing unit tests

PR #304, opened 2024-11-26
config: add path validation

ffromani:
add validation and restrictions about configuration paths,
to reduce the chance of path traversal vulnerabilities.

Most notably:
- config root path must be one of the allowed subpaths
- configlet are expected not to escape the configuration
  directories

### DIFF
--- a/pkg/config/validation.go
+++ b/pkg/config/validation.go
@@ -17,6 +17,12 @@ limitations under the License.
 package config
 
 import (
+	"errors"
+	"fmt"
+	"io/fs"
+	"path/filepath"
+	"strings"
+
 	metricssrv "github.com/k8stopologyawareschedwg/resource-topology-exporter/pkg/metrics/server"
 	"github.com/k8stopologyawareschedwg/resource-topology-exporter/pkg/resourcemonitor"
 )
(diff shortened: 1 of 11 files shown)


pkg/config/validation.go
```diff
@@ -36,3 +42,67 @@ func Validate(pArgs *ProgArgs) error {
 
 	return nil
 }
+
+func validateConfigLetPath(configletDir, configletName string) (string, error) {
+	configletPath, err := filepath.EvalSymlinks(filepath.Clean(filepath.Join(configletDir, configletName)))
```
Tal-or, line 47: `EvalSymlinks` failed if file doesn't exist.  
Could this be problematic for testing?

pkg/config/validation.go
```diff
@@ -36,3 +42,67 @@ func Validate(pArgs *ProgArgs) error {
...
+	if err != nil {
+		return "", err
+	}
+	if filepath.Dir(configletPath) != configletDir {
+		return "", fmt.Errorf("configlet %q is not within %q", configletName, configletDir)
```
Tal-or, line 52: This is possible only in case of softlinks, so maybe we should imply that in the error.

pkg/config/validation.go
```diff
@@ -36,3 +42,67 @@ func Validate(pArgs *ProgArgs) error {
...
+	if pattern == "" {
+		return "", errors.New("failed to validate configRoot path: not matches any allowed pattern")
+	}
+	return cfgRoot, nil
+}
+
+// IsConfigRootAllowed checks if an *already cleaned and canonicalized* path is among the allowed list.
+// use `addDirFns` to inject user-dependent paths (e.g. $HOME). Returns the matched pattern, if any,
+// and error describing the failure. The error is only relevant if failed to inject user-provided paths.
+func IsConfigRootAllowed(cfgPath string, addDirFns ...func() (string, error)) (string, error) {
```
Tal-or, line 87: `addDirFns` is neat

pkg/config/validation.go
ffromani: yes, this is the reason for https://github.com/k8stopologyawareschedwg/resource-topology-exporter/pull/304/files#diff-27fc28a543a4903aa873a125029a1f3ae2833f08470eae1c3c64b55e222810acR143

pkg/config/validation.go
ffromani: thanks :)

PR #318, opened 2025-04-08
deps: bump to kube 1.32, ghw 0.16

ffromani:
bump the required golang to 1.23, and bump kube deps (and related, like controller-runtime) to 1.32.
Finally, also bump `ghw` to 0.16.


shajmakh:
Thanks!
/lgtm
/approve

PR #336, opened 2025-07-01
add ncat to the resource topology exporter image

rbaturov:
This is needed for e2e coverage testing network policies.


ffromani (approved):
the value here is that the sigs.k8s.io scheduler plugin images also ship `nc`, so we conform to make e2e testing easier. I see no major problems here

PR #349, opened 2025-08-11
nrtupdater: add daemonset ownerReference

shajmakh:
We want to have an ownerReference added to NRT to point to the daemonset owner so that when the daemonset is deleted the NRT is deleted as well.

Note: if both node and daemonset owners are set, cleaning up the NRT automatically after deleting dependent objects will require deletion of **all** the owners.

### DIFF
--- a/pkg/config/flags.go
+++ b/pkg/config/flags.go
@@ -77,7 +77,7 @@ func FromFlags(pArgs *ProgArgs, args ...string) (string, string, error) {
 	CommandLine.StringVar(&pArgs.RTE.MetricsTLSCfg.KeyFile, "metrics-key-file", pArgs.RTE.MetricsTLSCfg.KeyFile, "key file name for TLS metrics serving")
 	CommandLine.BoolVar(&pArgs.RTE.MetricsTLSCfg.WantCliAuth, "metrics-want-cli-auth", pArgs.RTE.MetricsTLSCfg.WantCliAuth, "Toggle if client certificate and authentication is required")
 
-	CommandLine.StringVar(&refCnt, "reference-container", pArgs.RTE.ReferenceContainer.String(), "Reference container, used to learn about the shared cpu pool\n See: https://github.com/kubernetes/kubernetes/issues/102190\n format of spec is namespace/podname/containername.\n Alternatively, you can use the env vars REFERENCE_NAMESPACE, REFERENCE_POD_NAME, REFERENCE_CONTAINER_NAME.")
+	CommandLine.StringVar(&refCnt, "reference-container", pArgs.RTE.ReferenceContainer.String(), "Reference container, used to learn about the shared cpu pool\n See: https://github.com/kubernetes/kubernetes/issues/102190\n format of spec is namespace/podname/containername.\n Alternatively, you can use the env vars REFERENCE_NAMESPACE, REFERENCE_UID, REFERENCE_POD_NAME, REFERENCE_CONTAINER_NAME.")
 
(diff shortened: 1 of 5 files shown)


pkg/config/flags.go
```diff
@@ -77,7 +77,7 @@ func FromFlags(pArgs *ProgArgs, args ...string) (string, string, error) {
 	CommandLine.StringVar(&pArgs.RTE.MetricsTLSCfg.KeyFile, "metrics-key-file", pArgs.RTE.MetricsTLSCfg.KeyFile, "key file name for TLS metrics serving")
 	CommandLine.BoolVar(&pArgs.RTE.MetricsTLSCfg.WantCliAuth, "metrics-want-cli-auth", pArgs.RTE.MetricsTLSCfg.WantCliAuth, "Toggle if client certificate and authentication is required")
 
-	CommandLine.StringVar(&refCnt, "reference-container", pArgs.RTE.ReferenceContainer.String(), "Reference container, used to learn about the shared cpu pool\n See: https://github.com/kubernetes/kubernetes/issues/102190\n format of spec is namespace/podname/containername.\n Alternatively, you can use the env vars REFERENCE_NAMESPACE, REFERENCE_POD_NAME, REFERENCE_CONTAINER_NAME.")
+	CommandLine.StringVar(&refCnt, "reference-container", pArgs.RTE.ReferenceContainer.String(), "Reference container, used to learn about the shared cpu pool\n See: https://github.com/kubernetes/kubernetes/issues/102190\n format of spec is namespace/podname/containername.\n Alternatively, you can use the env vars REFERENCE_NAMESPACE, REFERENCE_UID, REFERENCE_POD_NAME, REFERENCE_CONTAINER_NAME.")
```
ffromani, line 80: we should not conflate the reference UID here, it's a different concept used in a similar but different way

pkg/nrtupdater/nrtupdater.go
```diff
@@ -176,6 +178,24 @@ func (te *NRTUpdater) updateNRTInfo(nrt *v1alpha2.NodeResourceTopology, info Mon
 // Check nrt.OwnerReferences for Node references and update it so it has only one Node reference,
 // the one to the Node with the same name as the NRT.
 func (te *NRTUpdater) updateOwnerReferences(nrt *v1alpha2.NodeResourceTopology) {
+	addDaemonSetOwner(nrt)
```
ffromani, line 181: I think we should add the daemonset owner
1. only if explicitly enabled
2. only if the node ownership is explicitly disabled

the most correct thing to do is the node ownership. From a scheduler perspective, this (using the node ownership as main one) enables the transparent replacement RTE <=> NFD.

pkg/nrtupdater/nrtupdater.go
```diff
@@ -191,18 +211,58 @@ func (te *NRTUpdater) updateOwnerReferences(nrt *v1alpha2.NodeResourceTopology)
...
-		{
-			Name:  "topologyManagerPolicy",
-			Value: te.tmConfig.Policy,
-		},
+func addDaemonSetOwner(nrt *v1alpha2.NodeResourceTopology) {
+	name, ok := os.LookupEnv("REFERENCE_POD_NAME")
+	if !ok {
+		resetOwnerReferences(nrt, "DaemonSet")
+		return
+	}
```
ffromani, line 222: why is not the UID sufficient?

ffromani (commented):
I was puzzled at first, but this is a nice and interesting approach. We should probably fine tune the implementation but I grew to like this direction. Nice!

shajmakh:
turns out this is a bad way to address the current problem of manual garbage collection taken into considiration the cost of this on related consumers of RTEs pubulated data.

PR #363, opened 2025-09-24
resourcemonitor: use `record` pkg to hold PFPStatus

shajmakh:
Use the new package record to store the PFP status into a an object that would hold all recorded statuses. At the meantime, this is configured to hold the last PFPStatus.


shajmakh:
wait for https://github.com/openshift-kni/debug-tools/pull/94

go.mod
```diff
@@ -11,22 +11,23 @@ require (
 	github.com/jaypipes/ghw v0.16.0
 	github.com/jeremywohl/flatten/v2 v2.0.0-20211013061545-07e4a09fb8e4
 	github.com/k8stopologyawareschedwg/noderesourcetopology-api v0.1.1
-	github.com/k8stopologyawareschedwg/podfingerprint v0.2.2
+	github.com/k8stopologyawareschedwg/podfingerprint v0.2.3
 	github.com/mdomke/git-semver v1.0.0
 	github.com/onsi/ginkgo/v2 v2.22.0
 	github.com/onsi/gomega v1.36.1
+	github.com/openshift-kni/debug-tools v0.2.5
```
ffromani, line 18: ATM I'm not sold about adding this dependency. This project should depend only on community backed/vendor neutral packages.

ffromani:
I'd rather decouple the functionalities and let consumer of the RTE packages implement their solution: https://github.com/k8stopologyawareschedwg/resource-topology-exporter/pull/364

go.mod
shajmakh: I see your point, thanks for pointing out. In this case this PR no longer needed

PR #367, opened 2026-01-14
config: narrow down dump-config options

ffromani:
remove rarely used and potentially (very much in theory) dump config options, considering dump config is in turn rarely used.

- drop `.andexit`. Use this pattern instead:
```
rte --dump-config=- --version
```
- drop custom file target. Use stdout redirect instead.


Tal-or (approved):
/lgtm
/approve

PR #389, opened 2026-03-26
deps: bump to kube 0.35.z

ffromani:
includes transitive dependencies like controller-runtime, which was previously misaligned.


swatisehgal (approved):
/lgtm

PR #390, opened 2026-04-03
attr: encode containers' numa placements

shajmakh:
Populate numaplacement.AttributeMetadata from encoded container affinities
alongside existing pod fingerprint attributes when enabled.

Please see the commits for detailed steps to enable this.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -174,6 +174,8 @@ func Finalize(pArgs *ProgArgs) error {
 		}
 	}
 
+	pArgs.Resourcemonitor.TopologyManagerPolicy = pArgs.RTE.TopologyManagerPolicy
+
 	return err
 }
 
--- a/pkg/podres/filter/numalocality/numalocality.go
+++ b/pkg/podres/filter/numalocality/numalocality.go
@@ -17,6 +17,8 @@ limitations under the License.
 package numalocality
 
 import (
+	"fmt"
+
 	podresourcesapi "k8s.io/kubelet/pkg/apis/podresources/v1"
 
 	podresfilter "github.com/k8stopologyawareschedwg/resource-topology-exporter/pkg/podres/filter"
@@ -109,4 +111,45 @@ func VerifyContainer(cnt *podresourcesapi.ContainerResources) podresfilter.Resul
 	return podresfilter.Result{
 		Allow: false,
 	}
+}
+
+// ResolveContainerPlacement finds the single NUMA node placement for a container;
+// it returns the NUMA node ID if found, otherwise it returns -1.
+// IMPORTANT: multiple-NUMA affinity is not supported (yet), thus this should be called only
+// on single NUMA node topology manager policy.
+func ResolveContainerPlacement(coreIDToNodeIDMap map[int]int, cnt *podresourcesapi.ContainerResources) (bool, int, error) {
+	eligibleForPlacement := true
+	if cnt == nil {
+		return !eligibleForPlacement, -1, fmt.Errorf("nil container resources")
+	}
+
+	if len(cnt.CpuIds) > 0 {
+		nodeID, ok := coreIDToNodeIDMap[int(cnt.CpuIds[0])]
+		if !ok {
+			//should never happen
+			return eligibleForPlacement, -1, fmt.Errorf("CPU ID %d not found in coreIDToNodeIDMap", cnt.CpuIds[0])
+		}
+		return eligibleForPlacement, nodeID, nil
+	}
+
+	// TODO: handle multi-NUMAs for devices
+	// we need to know if on multi-NUMA topology this data is available by
+	// the pod-resources API or if it needs other means to get the NUMA node ID
+	// https://redhat.atlassian.net/browse/CNF-23537
+	// currently this assumes single NUMA node for all devices containers
+	for _, dev := range cnt.Devices {
+		nodeID := GetNUMAID(dev.Topology)
+		if len(dev.DeviceIds) > 0 && nodeID != -1 {
+			return eligibleForPlacement, nodeID, nil
+		}
+	}
+
+	for _, mem := range cnt.Memory {
+		nodeID := GetNUMAID(mem.Topology)
+		if nodeID != -1 {
+			return eligibleForPlacement, nodeID, nil
+		}
+	}
+
+	return !eligibleForPlacement, -1, nil
 }
--- a/pkg/podres/filter/numalocality/numalocality.go
+++ b/pkg/podres/filter/numalocality/numalocality.go
@@ -56,18 +56,18 @@ func Required(pr *podresourcesapi.PodResources) bool {
 	return got.Allow
 }
 
-func IsPresent(topo *podresourcesapi.TopologyInfo) bool {
+func GetNUMAID(topo *podresourcesapi.TopologyInfo) int {
 	if topo == nil || topo.Nodes == nil {
-		return false
+		return -1
 	}
 	// if Nodes is not given, this means "don't care about locality". It's a legal representation.
 	for _, node := range topo.Nodes {
 		// setting node.ID == -1 is also a legal representation for "don't care about locality".
 		if node.ID >= 0 {
-			return true
+			return int(node.ID)
 		}
 	}
-	return false
+	return -1
 }
 
 func VerifyContainer(cnt *podresourcesapi.ContainerResources) podresfilter.Result {
@@ -88,7 +88,7 @@ func VerifyContainer(cnt *podresourcesapi.ContainerResources) podresfilter.Resul
 		}
 	}
 	for _, mem := range cnt.Memory {
-		if IsPresent(mem.Topology) {
+		if GetNUMAID(mem.Topology) != -1 {
 			return podresfilter.Result{
 				Allow:  true,
 				Ident:  cnt.Name,
@@ -97,7 +97,7 @@ func VerifyContainer(cnt *podresourcesapi.ContainerResources) podresfilter.Resul
 		}
 	}
 	for _, dev := range cnt.Devices {
-		if len(dev.DeviceIds) > 0 && IsPresent(dev.Topology) {
+		if len(dev.DeviceIds) > 0 && GetNUMAID(dev.Topology) != -1 {
 			return podresfilter.Result{
 				Allow:  true,
 				Ident:  cnt.Name,
--- a/pkg/podres/filter/numalocality/numalocality.go
+++ b/pkg/podres/filter/numalocality/numalocality.go
@@ -35,33 +35,9 @@ func Verify(pr *podresourcesapi.PodResources) podresfilter.Result {
 		}
 	}
 	for _, cr := range pr.Containers {
-		// there's no correct order for checks here, or faster.
-		// CPUs are the most frequent (because there's always here) exclusively
-		// assigned devices, so we start from here.
-		if len(cr.CpuIds) > 0 {
-			return podresfilter.Result{
-				Allow:  true,
-				Ident:  cr.Name,
-				Reason: CPU,
-			}
-		}
-		for _, mem := range cr.Memory {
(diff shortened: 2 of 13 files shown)


pkg/resourcemonitor/resourcemonitor.go
```diff
@@ -322,6 +323,16 @@ func (rm *resourceMonitor) Scan(excludeList ResourceExclude) (ScanResponse, erro
 			Resources: make(topologyv1alpha2.ResourceInfoList, 0),
 		}
 
+		if payload.NUMANodes > 0 {
+			zoneVector, ok := payload.Vectors[nodeID]
+			if ok {
+				zone.Attributes = append(zone.Attributes, topologyv1alpha2.AttributeInfo{
+					Name:  numaplacement.AttributeVector,
+					Value: zoneVector,
+				})
```
shajmakh, line 332: I need to ensure this is not adding the attributes over and over again

pkg/config/config.go
```diff
@@ -174,6 +174,8 @@ func Finalize(pArgs *ProgArgs) error {
 		}
 	}
 
+	pArgs.Resourcemonitor.TopologyManagerPolicy = pArgs.RTE.TopologyManagerPolicy
```
ffromani, line 177: I'm not clear what this commits wants to do and this hunk in particular

pkg/resourcemonitor/resourcemonitor.go
```diff
@@ -52,6 +52,8 @@ import (
 const (
 	defaultPodResourcesTimeout = 10 * time.Second
 	// obtained these values from node e2e tests : https://github.com/kubernetes/kubernetes/blob/82baa26905c94398a0d19e1b1ecf54eb8acb6029/test/e2e_node/util.go#L70
+
+	TopologyManagerPolicySingleNUMANode = "single-numa-node"
```
ffromani, line 56: is this constant used?

pkg/resourcemonitor/resourcemonitor.go
```diff
@@ -75,6 +77,7 @@ type Args struct {
 	PodSetFingerprintStatusFile string          `json:"podSetFingerprintStatusFile,omitempty"`
 	PodExclude                  podexclude.List `json:"podExclude,omitempty"`
 	ExcludeTerminalPods         bool            `json:"excludeTerminalPods,omitempty"`
+	TopologyManagerPolicy       string          `json:"topologyManagerPolicy,omitempty"`
```
ffromani, line 80: why do we need this? it seems duplication

pkg/podres/filter/numalocality/numalocality.go
```diff
@@ -35,33 +35,9 @@ func Verify(pr *podresourcesapi.PodResources) podresfilter.Result {
...
-		}
-		for _, dev := range cr.Devices {
-			if len(dev.DeviceIds) > 0 && IsPresent(dev.Topology) {
-				return podresfilter.Result{
-					Allow:  true,
-					Ident:  cr.Name,
-					Reason: Device,
-				}
-			}
+		res := VerifyContainer(cr)
```
ffromani, line 38: extracting the function is fine, but we need to move out from this package. The "filter" subpackages are meant to be used as components, not as libraries. We need a new library package which both this filter and the new code will consume

pkg/podres/filter/numalocality/numalocality.go
```diff
@@ -93,3 +69,44 @@ func IsPresent(topo *podresourcesapi.TopologyInfo) bool {
 	}
 	return false
 }
+
+func VerifyContainer(cnt *podresourcesapi.ContainerResources) podresfilter.Result {
```
ffromani, line 73: VerifyContainer about what? `numalocality.VerifyContainer` is not really telling, and this is foing to change if we move (and we should) this code. Let's think about a better name

pkg/podres/filter/numalocality/numalocality.go
```diff
@@ -56,18 +56,18 @@ func Required(pr *podresourcesapi.PodResources) bool {
 	return got.Allow
 }
 
-func IsPresent(topo *podresourcesapi.TopologyInfo) bool {
+func GetNUMAID(topo *podresourcesapi.TopologyInfo) int {
 	if topo == nil || topo.Nodes == nil {
-		return false
+		return -1
```
ffromani, line 61: -1 doubles as "any numa node". Is the correct contract here or we need a different value for "unspecified/unknown"?

pkg/podres/filter/numalocality/numalocality.go
```diff
@@ -110,3 +112,44 @@ func VerifyContainer(cnt *podresourcesapi.ContainerResources) podresfilter.Resul
 		Allow: false,
 	}
 }
+
+// ResolveContainerPlacement finds the single NUMA node placement for a container;
+// it returns the NUMA node ID if found, otherwise it returns -1.
+// IMPORTANT: multiple-NUMA affinity is not supported (yet), thus this should be called only
+// on single NUMA node topology manager policy.
+func ResolveContainerPlacement(coreIDToNodeIDMap map[int]int, cnt *podresourcesapi.ContainerResources) (bool, int, error) {
```
ffromani, line 120: this does belong into a package dealing with numalocality, but doesn't belong into a filter, which is a terminal (consumer) package not a library package

pkg/podres/filter/numalocality/numalocality.go
```diff
@@ -110,3 +112,44 @@ func VerifyContainer(cnt *podresourcesapi.ContainerResources) podresfilter.Resul
...
+	eligibleForPlacement := true
+	if cnt == nil {
+		return !eligibleForPlacement, -1, fmt.Errorf("nil container resources")
+	}
+
+	if len(cnt.CpuIds) > 0 {
+		nodeID, ok := coreIDToNodeIDMap[int(cnt.CpuIds[0])]
```
ffromani, line 127: why only the first?

pkg/resourcemonitor/resourcemonitor.go
```diff
@@ -291,6 +293,19 @@ func (rm *resourceMonitor) Scan(excludeList ResourceExclude) (ScanResponse, erro
...
+		if err != nil {
+			klog.V(2).ErrorS(err, "resmon: failed to encode container affinities")
+		} else {
+			metadataAttValue = payload.PackMetadata()
+		}
+
+		scanRes.Attributes = append(scanRes.Attributes, topologyv1alpha2.AttributeInfo{
+			Name:  numaplacement.AttributeMetadata,
+			Value: metadataAttValue,
+		})
```
ffromani, line 308: we need to document that the numaplacement info depends on the PFP enablement. It is non-obvious.

Other than that, I'll rethink the loop in Scan(). We need to process all the pods which pass the `numalocality` check and make logic on these. I'd start inilining `computePodFingerprintFromPodResources` in the one and calling site and reorganizing the flow.

pkg/resourcemonitor/resourcemonitor.go
```diff
@@ -458,6 +473,45 @@ func (rm *resourceMonitor) updateNodeResources() error {
 	return nil
 }
 
+func (rm *resourceMonitor) computeNUMAPlacementPayload(podRes []*podresourcesapi.PodResources) (numaplacement.Payload, error) {
+	if rm.args.TopologyManagerPolicy != TopologyManagerPolicySingleNUMANode {
+		klog.Infof("resmon: topology manager policy is not single-numa-node, skipping container NUMA affinity encoding")
+		return numaplacement.Payload{}, nil
+	}
```
ffromani, line 480: this check probably makes sense but it should be moved earlier in the flow. And again I'd start inlining this function in its one and only calling site and then reorganize the flow

pkg/resourcemonitor/resourcemonitor.go
```diff
@@ -458,6 +473,45 @@ func (rm *resourceMonitor) updateNodeResources() error {
...
+
+	for _, pr := range podRes {
+		for _, cnt := range pr.Containers {
+			eligibleForPlacement, numaNodeID, err := numalocality.ResolveContainerPlacement(rm.coreIDToNodeIDMap, cnt)
+			if !eligibleForPlacement {
+				continue
+			}
+
+			if err != nil || numaNodeID == -1 {
```
ffromani, line 503: does it make sense to return `eligibleForPlacement=true` AND `numaNodeID=-1`? if not probably ResolveContainerPlacement can use a refactoring.

ffromani (commented):
Thanks for the contribution, this is an important feature. Let's make sure it's opt-in and backward compatible (preferably both)
partial update, covering the general architecture of the change

pkg/config/config.go
shajmakh: we need to publish the numaplacement attributes only when the node's topology policy is single-numa-node.

pkg/resourcemonitor/resourcemonitor.go
shajmakh: yes but in later commit:
https://github.com/k8stopologyawareschedwg/resource-topology-exporter/pull/390/changes/1910cd3c974c66f7bee888bb1d9582c7f7ff89a7#diff-00b22e6b5678bf5d89eea5e7fdc1409bb5974da363b42d5f0413716882ab9271R477
I should probably move this const to the commit it is first used in

pkg/podres/filter/numalocality/numalocality.go
shajmakh: it is the same as verifying the level above which is pod, like Verify() was to verify if a pod is eligible for numalocality if it had any container which is eligible, so I exported the specific code to verify a single container's numalocality. however I can see your point especially looking at the returned result word combination. I'll update this. Thanks for pointing it out.

pkg/podres/filter/numalocality/numalocality.go
shajmakh: there is a gap here indeed. we need to loop over all the cpus. I'll fix that and the returned result. 
However in the flow of RTE-numaplacement, as we only care to compute the payload only when the TMPolicy is single-numa-node, the first CPU would be enough. yet this specific library package should be improved and not coupled or dependent on the previous flow. This should actually be applied to all the container's resources too (devices, memory). Will handle that. Thanks for highlighting.

pkg/resourcemonitor/resourcemonitor.go
shajmakh: it makes sense if it was found that a container's resources are pinned to multiple NUMAs, which is a gap here that is not handled. this is a calling for refactoring indeed.

pkg/resourcemonitor/resourcemonitor.go
shajmakh: got it, thanks for your input, will address this.

pkg/resourcemonitor/resourcemonitor.go
shajmakh: +1 on the doc, will address that. 
regarding inlining `computePodFingerprintFromPodResources` I'm against that because then how will we test the payload computation unit then? I agree this is used only in one place, but if anything is messed up during the computation, it will be hard to troubleshoot the exact breaking point. with the function in place, we can have 100% control and test coverage on the flow, which is why I prefer to break the readability flow over having a complicated test matrix. let me know WDYT and if I misunderstood your point here please.

ffromani: I think I expressed myself poorly. The proposal is not just inline and call it a day: is it inline, have all the new flows coherently in the calling function, and then revisit and perform a new targeted refactoring starting from the full scope.

pkg/config/config.go
ffromani: yes, we talked offline, and the producer side should only populate data if the node configuration is such as enforcing actual numa locality (which depends indeed on the topology manager policy). The consumer side (scheduler) MUST ignore numalocality data if the node doesn't enforce so, but still no point in producing this data in the first place. Point taken, you are correct here.

PR #391, opened 2026-04-17
podreadiness: optimize condition injection

ffromani:
At steady state, we can observe 2 pod GETs per update cycle due to condition injection.
This is actually unnecessary: if the conditions we computed didn't change, we should not send updates.

The RTE is the sole owner of these specific conditions, so it is authoritative on them. OTOH, we now open the door to tampering, which the previous code prevented. But at scale, the API server load is halved, which is worth the tiny risk.

Should we need to be super safe, we can add an option to revert to the old behavior, but at this stage it seems unnecessary.

### DIFF
--- a/pkg/podreadiness/conditioninjector.go
+++ b/pkg/podreadiness/conditioninjector.go
@@ -35,9 +35,9 @@ func NewConditionInjector(cs kubernetes.Interface) (*ConditionInjector, error) {
 	}, nil
 }
 
-func (ci *ConditionInjector) Inject(cond v1.PodCondition) error {
+func (ci *ConditionInjector) Inject(ctx context.Context, cond v1.PodCondition) error {
 	conditionExist := false
-	pod, err := ci.cs.CoreV1().Pods(ci.ns).Get(context.TODO(), ci.podName, metav1.GetOptions{})
+	pod, err := ci.cs.CoreV1().Pods(ci.ns).Get(ctx, ci.podName, metav1.GetOptions{})
 	if err != nil {
 		return err
 	}
@@ -62,18 +62,20 @@ func (ci *ConditionInjector) Inject(cond v1.PodCondition) error {
 
 	klog.V(4).Infof("pod conditions: %v", pod.Status.Conditions)
 
-	_, err = ci.cs.CoreV1().Pods(ci.ns).UpdateStatus(context.TODO(), pod, metav1.UpdateOptions{})
+	_, err = ci.cs.CoreV1().Pods(ci.ns).UpdateStatus(ctx, pod, metav1.UpdateOptions{})
 	return err
 }
 
-func (ci *ConditionInjector) Run(condChan <-chan v1.PodCondition) {
-	go func() {
-		for {
-			cond := <-condChan
-			err := ci.Inject(cond)
+func (ci *ConditionInjector) Run(ctx context.Context, condChan <-chan v1.PodCondition) {
+	for {
+		select {
+		case cond := <-condChan:
+			err := ci.Inject(ctx, cond)
 			if err != nil {
 				klog.Errorf("failed to update pod status with condition: %v", cond)
 			}
+		case <-ctx.Done():
+			return
 		}
-	}()
+	}
 }
--- a/pkg/podreadiness/conditioninjector.go
+++ b/pkg/podreadiness/conditioninjector.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"slices"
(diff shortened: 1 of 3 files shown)


ffromani:
@Tal-or @shajmakh

pkg/podreadiness/conditioninjector.go
```diff
@@ -62,18 +62,24 @@ func (ci *ConditionInjector) Inject(cond v1.PodCondition) error {
...
 	klog.V(4).Infof("pod conditions: %v", pod.Status.Conditions)
 
-	_, err = ci.cs.CoreV1().Pods(ci.ns).UpdateStatus(context.TODO(), pod, metav1.UpdateOptions{})
-	return err
+	_, err = ci.cs.CoreV1().Pods(ci.ns).UpdateStatus(ctx, pod, metav1.UpdateOptions{})
+	if err != nil {
+		return err
+	}
+
+	return nil
```
shajmakh, line 76: should have stayed `return err`?

pkg/podreadiness/conditioninjector.go
```diff
@@ -27,15 +37,15 @@ func NewConditionInjector(cs kubernetes.Interface) (*ConditionInjector, error) {
...
-		podName: podVal,
-	}, nil
+	return NewConditionInjectorWithIdentity(cs, nsVal, podVal), nil
 }
 
 func (ci *ConditionInjector) Inject(ctx context.Context, cond v1.PodCondition) error {
+	if oldCond, ok := ci.lastUpdate[cond.Type]; ok && equalPodCondition(oldCond, cond) {
+		// avoid APIServer round trips if nothing changed
+		return nil
+	}
```
shajmakh, line 48: is there a chance that the cached status be stale and that here it compares against a no-longer-valid condition? if that is the case then I think this will disallow to update the pod

pkg/podreadiness/conditioninjector.go
```diff
@@ -67,6 +79,7 @@ func (ci *ConditionInjector) Inject(ctx context.Context, cond v1.PodCondition) e
 		return err
 	}
 
+	ci.lastUpdate[cond.Type] = cond
```
shajmakh, line 75: correct me if mistaken, so this is the only method that updates the cache, right?

pkg/podreadiness/conditioninjector.go
shajmakh: my concern is what if this condition passed (old vs new is the same) while in lines https://github.com/k8stopologyawareschedwg/resource-topology-exporter/pull/391/changes/8f8a8ec2771d89c1d32bbf4b6048f48f0ae01a61#diff-69b3cd770ed4e578b43a7629a8c8f275966b82b54a1e63fa0bd85b26cc3356dcR50-R51 the pod conditions did change

pkg/podreadiness/conditioninjector.go
```diff
@@ -45,8 +55,10 @@ func (ci *ConditionInjector) Inject(ctx context.Context, cond v1.PodCondition) e
 	for pos, podCond := range pod.Status.Conditions {
 		if podCond.Type == cond.Type {
 			conditionExist = true
-			if podCond.Status == cond.Status {
+			if equalPodCondition(podCond, cond) {
 				// do nothing condition already updated
+				klog.V(4).Infof("pod conditions already set as expected")
+				ci.lastUpdate[cond.Type] = cond
```
shajmakh: why again if we trust the cache check above? this exactly raises my question above, it doesn't seem that we are sure that the current cache already have the real value

shajmakh (commented):
Thanks for this, I see you addressed a memory leak in improving Run, looks good. left few comments.

pkg/podreadiness/conditioninjector.go
ffromani: yes, indeed.

pkg/podreadiness/conditioninjector.go
ffromani: yes, but we handle 2 conditions

pkg/podreadiness/conditioninjector.go
ffromani: good point, simplified the code to make it consistent

pkg/podreadiness/conditioninjector.go
shajmakh: we actually needed that for the second commit. NVM

shajmakh:
/lgtm
Thanks!

shajmakh (approved):

PR #392, opened 2026-04-20
introduce patch support to optimize nrt update

ffromani:
We can use the Patch verb instead of plain get+update to reduce API calls and also wire traffic.
API calls go down 2->1;
Wire traffic goes down at least 33%, possibly 50% in the best (and expected) scenario.

We add metric to detect the patch sizes, which determine the wire load traffic; the closer the patch size to 100%, the worse the wire traffic gets and the savings trend down towards 33%, which is the guaranteed improvement:

```
GET + UPDATE flow (`sendObjectUpdate`)
GET request (client -> server) ~negligible (object name only)
GET response (server -> client) full NRT object
PUT request (client -> server) full NRT object (mutated)
PUT response (server -> client) full NRT object (as stored)
Total: ~3x full object
```

```
PATCH flow (`patchNRT`)
PATCH request (client -> server) JSON merge patch (delta: worst case
full NRT object)
PATCH response (server -> client) full NRT object (as stored)
Total: 1x full object + delta best case, 2x full object worst case
```

In order to fight back the NRT object drift, we add an option to force the internal resync falling back to get+update every N user-provided cycles. Note the object drift will cause patch size increase, because the k8s metadata must never surface on the patch itself.

Finally, the patch mode has a built-in safety fallback on the get+update flow should the patch fail to be generate or apply.


pkg/metrics/metrics.go
```diff
@@ -56,7 +67,7 @@ func UpdateNodeResourceTopologyWritesMetric(operation, trigger string) {
 	}).Inc()
 }
 
-func UpdatePodResourceApiCallsFailureMetric(funcName string) {
+func UpdatePodResourceApiCallsFailuresMetric(funcName string) {
```
shajmakh, line 70: nit: API

pkg/nrtupdater/nrtupdater.go
```diff
@@ -139,8 +146,88 @@ func (te *NRTUpdater) sendData(ctx context.Context, cli topologyclientset.Interf
...
+}
+
+func (pi NRTPatchInfo) SizeRatio() float64 {
+	if pi.FullObjBytes == 0 {
+		return 1.0
+	}
+	return float64(len(pi.Patch)) / float64(pi.FullObjBytes)
+}
+
+func MakeNRTPatch(nrtOld, nrtNew *v1alpha2.NodeResourceTopology) (NRTPatchInfo, string, error) {
```
shajmakh, line 164: I'm thinking if we need to couple the failure result into a single struct including `reason` and `err` given that the reason s empty only on nil error. not a big deal though.

pkg/nrtupdater/nrtupdater.go
```diff
@@ -217,16 +218,30 @@ func (te *NRTUpdater) patchNRT(ctx context.Context, cli topologyclientset.Interf
...
+		nrtObj, err := te.patchNRT(ctx, cli, info)
+		if err == nil {
+			te.patchCount++
+			return nrtObj, nil
 		}
-		te.prevNRT = nrtObj
-		return nrtObj, nil
+	} else {
+		klog.V(2).Infof("nrtupdater forcing resync after %d patches", te.patchCount)
+	}
```
shajmakh, line 239: we can remove the else here and just simply log the info about proceeding with Update

manifests/resource-topology-exporter-config.yaml
```diff
@@ -24,6 +24,9 @@ data:
     global:
       verbose: 6
       debug: true
+    nrtUpdater:
+      patchMode: false
+      patchResync: 10
```
shajmakh: shouldn't patchMode be true? IIUC it wouldn't make sense to set the resync counter without actually enable the patch mode. Also, isn't the intention of all this to optimize API update calls, so why not enable that by default. did I miss something?

shajmakh (commented):
Good to learn from this. Thanks for filing the PR.
Overall LGTM except that I see that the patch mode is not meant to be the default. Is that intended?

pkg/metrics/metrics.go
ffromani: renaming Api -> API is a good idea which I think is better done on a separate PR

pkg/nrtupdater/nrtupdater.go
ffromani: `reason` and `err` are coupled indeed, but neither of them is really belonging to a `NRTPatchInfo`. I'd prefer to keep them split for now. Perhaps the best way forward is to rethink the return value shape entirely, but that's a separate effort

pkg/nrtupdater/nrtupdater.go
ffromani: as much as I dislike the `else`s, in this case we really want to do this iff the patchResync count mandates so, otherwise we are adding confusing noise to the logs.

manifests/resource-topology-exporter-config.yaml
ffromani: Yes, we can turn it by default in 0.25 and onwards. Backports, if ever, would however have  default to off for backward compatibility.

pkg/nrtupdater/nrtupdater.go
shajmakh: okay right, makes sense

manifests/resource-topology-exporter-config.yaml
shajmakh: > Backports, if ever, would however have default to off for backward compatibility.

right because of the RBACs. almost forgot about these.

shajmakh (approved):

PR #393, opened 2026-04-23
config: prefer config files over command line flags

ffromani:
Even though they will still be fully supported for the foreseeable future, we should use configfiles as default config mechanism. 

Update the example configuration.
This move highlighted some gaps in the config flags/file mapping, which are fixed in the same commit.

### DIFF
--- a/pkg/config/cfgfile_test.go
+++ b/pkg/config/cfgfile_test.go
@@ -68,6 +68,37 @@ func TestReadResourceExclude(t *testing.T) {
 	}
 }
 
+func TestMetricsAddressFromConfig(t *testing.T) {
+	testDir, closer := setupTest(t)
+	t.Cleanup(closer)
+
+	confRoot := filepath.Join(testDir, "test-metrics-address")
+	daemonDir := filepath.Join(confRoot, "daemon")
+	if err := os.MkdirAll(daemonDir, 0755); err != nil {
+		t.Fatalf("unexpected error creating daemon dir: %v", err)
+	}
+
+	expected := `topologyExporter:
+  metricsAddress: 192.168.1.100
+`
+	expectedAddress := "192.168.1.100"
+
+	if err := os.WriteFile(filepath.Join(daemonDir, "config.yaml"), []byte(expected), 0644); err != nil {
+		t.Fatalf("unexpected error writing config: %v", err)
+	}
+
+	var pArgs ProgArgs
+	SetDefaults(&pArgs)
+	extraPath := FixExtraConfigPath(confRoot)
+	if err := FromFiles(&pArgs, confRoot, extraPath); err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+
+	if pArgs.RTE.MetricsAddress != expectedAddress {
+		t.Errorf("MetricsAddress: got %q expected %q", pArgs.RTE.MetricsAddress, expectedAddress)
+	}
+}
+
 func TestFromFiles(t *testing.T) {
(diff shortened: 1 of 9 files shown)


pkg/config/cfgfile_test.go
```diff
@@ -68,6 +68,36 @@ func TestReadResourceExclude(t *testing.T) {
...
+	t.Cleanup(closer)
+
+	confRoot := filepath.Join(testDir, "test-metrics-address")
+	daemonDir := filepath.Join(confRoot, "daemon")
+	if err := os.MkdirAll(daemonDir, 0755); err != nil {
+		t.Fatalf("unexpected error creating daemon dir: %v", err)
+	}
+
+	cfgContent := `topologyExporter:
+  metricsAddress: 192.168.1.100
```
shajmakh, line 82: nit: use the variable expected here instead.

ffromani: thanks, done

test/e2e/rte/rte.go
```diff
@@ -97,63 +98,60 @@ var _ = ginkgo.Describe("[RTE][InfraConsuming] Resource topology exporter", func
...
+				var err error
+				baselineNodeTopo, err = f.TopoCli.TopologyV1alpha2().NodeResourceTopologies().Get(context.TODO(), topologyUpdaterNode.Name, metav1.GetOptions{})
+				if err != nil {
+					return fmt.Errorf("failed to get the node topology resource: %w", err)
+				}
+				if baselineNodeTopo.ObjectMeta.ResourceVersion == initialNodeTopo.ObjectMeta.ResourceVersion {
+					return fmt.Errorf("resource %s not yet updated - resource version not bumped", topologyUpdaterNode.Name)
+				}
+				klog.Infof("resource %s baseline update (resource version %v -> %v)", topologyUpdaterNode.Name, initialNodeTopo.ObjectMeta.ResourceVersion, baselineNodeTopo.ObjectMeta.ResourceVersion)
+				return nil
```
shajmakh: can use the gomega assertions here instead of returning an error.

ffromani: done, for all the e2e tests. Thanks!

test/e2e/rte/rte.go
```diff
@@ -272,18 +256,19 @@ var _ = ginkgo.Describe("[RTE][InfraConsuming] Resource topology exporter", func
...
-			gomega.Eventually(func() bool {
+			gomega.Eventually(func(g gomega.Gomega) {
 				nrt := e2enodetopology.GetNodeTopology(f.TopoCli, topologyUpdaterNode.Name)
 				devName := e2etestenv.GetDeviceName()
+				found := false
 				for _, zone := range nrt.Zones {
 					for _, res := range zone.Resources {
 						if res.Name == devName {
-							return true
+							found = true
```
shajmakh, line 266: no need to change but only for future similar cases, this is one example for when the return approach is better than gomega inner assertion. it is a simple single condition, when found is first assigned true the eventually returns and succeeds. whereas the current state requires to break the loops when found is set to true.

shajmakh (approved):
/lgtm
Thanks!

test/e2e/rte/rte.go
ffromani: using `func() bool` is the weakest option of all because makes it unnecessarily hard to build a good failure message. In this case, since we are improving tests, the best options are using `func() error` as closure or switch to `func(g Gomega()`. Picked up the latter for consistency.

PR #394, opened 2026-05-11
better support for multiple-instances on the same node

ffromani:
In order to support upgrade flows, turns out we actually need to support a temporary but non-zero window on which more than a RTE is  running on a node.

This requirement breaks some very key and very old design assumptions, but the saving grace is that RTEs are mostly stateless, all the instance observe the same state and patch mode is the default but not part of a stable release yet.

Add better support for this very scenario: more resilient patch mode, handling confilicts, and node-local inter-RTE basic lease. We don't use proper battle-tested kubernetes leases because the scenario we guard against is expected to be limited and time-bound, so a simpler solution has merit, and reduces the apiserver load.

### DIFF
--- a/pkg/nodelease/nodelease.go
+++ b/pkg/nodelease/nodelease.go
@@ -1,0 +1,93 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package nodelease
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"syscall"
+
+	"k8s.io/klog/v2"
+)
+
+const LeaseFileName = "lease"
+
+const AutodetectLeaseFromNotify = "default"
+
+type Lease interface {
+	TryLock() bool
+	Close() error
+}
+
+type flockLease struct {
+	fd   int
+	path string
+	held bool
+}
+
+type nullLease struct{}
+
+// NewNull returns a no-op lease that always succeeds.
+func NewNull() Lease {
+	return &nullLease{}
+}
+
+func (hl *nullLease) TryLock() bool { return true }
+func (hl *nullLease) Close() error  { return nil }
+
+// New opens (or creates) the lease file at the given path.
+// The lock is NOT acquired until TryLock is called.
+func New(leaseFilePath string) (Lease, error) {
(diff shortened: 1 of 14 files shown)


pkg/nodelease/nodelease.go
```diff
@@ -0,0 +1,93 @@
+// SPDX-License-Identifier: Apache-2.0
```
Tal-or, line 1: Do we have to write the lease logic by ourselves? 
Did we try to see if there are already working well-maintained libraries out there that provides such functionality?

pkg/resourcetopologyexporter/resourceobserver.go
```diff
@@ -78,6 +81,13 @@ func (rm *ResourceObserver) Run(eventsChan <-chan notification.Event, condChan c
 				podreadiness.SetCondition(condChan, podreadiness.PodresourcesFetched, condStatus)
 				continue
 			}
+
```
Tal-or, line 84: Maybe move this check to the beginning of the `case` to save some CPU cycles (the scanning work) in case it didn't manage to acquire the lock.

pkg/resourcetopologyexporter/resourcetopologyexporter.go
```diff
@@ -176,3 +185,17 @@ func getTopologyManagerSettings(rteArgs Args) (tmSettings, error) {
...
+
+func createNodeLease(rteArgs Args) (nodelease.Lease, error) {
+	if rteArgs.LeaseFilePath == "" {
+		return nodelease.NewNull(), nil
+	}
+	leaseFile := rteArgs.LeaseFilePath
+	if leaseFile == nodelease.AutodetectLeaseFromNotify {
+		leaseFile = nodelease.LeaseFilePath(rteArgs.NotifyFilePath)
+		if leaseFile == "" {
+			return nil, fmt.Errorf("lease file %q requested but no notify file path is set (set --notify-file or provide an explicit --lease-file path)", nodelease.AutodetectLeaseFromNotify)
```
Tal-or, line 197: Couldn't figure out why the notify-file and the lease-file are connected.
Worth adding a comment about that.

PR #395, opened 2026-05-12
config: ux/devex improvements

ffromani:
polish the `config` package with quality of life improvements. No behavior changes.

### DIFF
--- a/pkg/config/cfgfile.go
+++ b/pkg/config/cfgfile.go
@@ -80,7 +80,7 @@ func fromDaemonFiles(pArgs *ProgArgs, configPathRoot string) error {
 	// 1. we need to distinguish "value not set" from "value with defaults" (language's or our's)
 	//    To do so, using the simple unmashalling is not good enough, because unset values
 	//    will be set to zero values. But we want to LACK unset values, so we know we don't
-	//    apply over the previos iteration, keeping the oldest set value (possibly the default)
+	//    apply over the previous iteration, keeping the latest set value (possibly the default).
 	// 2. hence, we don't use the simple unmarshalling, but we unmarshal to generic maps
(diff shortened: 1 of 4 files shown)


pkg/config/cfgfile.go
```diff
@@ -80,7 +80,7 @@ func fromDaemonFiles(pArgs *ProgArgs, configPathRoot string) error {
 	// 1. we need to distinguish "value not set" from "value with defaults" (language's or our's)
 	//    To do so, using the simple unmashalling is not good enough, because unset values
 	//    will be set to zero values. But we want to LACK unset values, so we know we don't
-	//    apply over the previos iteration, keeping the oldest set value (possibly the default)
+	//    apply over the previous iteration, keeping the latest set value
```
shajmakh: I'd just append "(the default)" if that makes sense for a clearer expectation

pkg/config/environ.go
```diff
@@ -76,4 +80,30 @@ func FromEnv(pArgs *ProgArgs) {
 	if pArgs.Resourcemonitor.PodSetFingerprintStatusFile == "" {
 		pArgs.Resourcemonitor.PodSetFingerprintStatusFile = PodSetFingerprintStatusFileFromEnv()
 	}
+
+	logEnvOverrides(before, *pArgs)
```
shajmakh: it looks clean this way but is possible to miss for new env updates or deletions. how about we keep the logs near the actual override?

pkg/config/cfgfile.go
ffromani: thanks, fixed

pkg/config/environ.go
ffromani: sur,e done.

shajmakh (approved):

PR #396, opened 2026-05-26
resmon: encode containers' numa placements as attributes

shajmakh:
RTE entry point has access to podresourcesAPI from which we can gather
data about specific NUMA placement for containers. Scan() already
process the numa eligible pod resources but stops at the first
container. The new feature we desire to enable is to encode every
numa-eligible container and publish that data as an encoded attribute so
it would be used by the consumer.

The way the encoder works is LEB89 based and thus is optimized to N-1
instead of N. This requires that both RTE -being the producer- and the
consumer to be in sync in terms of the pods being processed; and that is
already handled by PFP. when PFP is in sync on the consumer end,
decoding the attributes is safe and reliable.


pkg/numalocality/numalocality.go
```diff
@@ -0,0 +1,93 @@
+package numalocality
```
ffromani, line 1: `pkg/podres/filter/...` is not the right place for library code. This tree is more akin a plugin than a library. We may have another library package which is consumed everywhere needed

pkg/resourcemonitor/resourcemonitor.go
```diff
@@ -173,6 +182,7 @@ func (nrc perNUMAResourceCounter) String() string {
 type resourceMonitor struct {
 	nodeName          string
 	args              Args
+	tmconf            nrtupdater.TMConfig
```
ffromani: this is a smell. This struct should not depend on other packages config structs.

pkg/resourcemonitor/resourcemonitor.go
```diff
@@ -458,31 +577,30 @@ func (rm *resourceMonitor) updateNodeResources() error {
 func computePodFingerprintFromPodResources(podRes []*podresourcesapi.PodResources, st *podfingerprint.Status, verifyFunc func(*podresourcesapi.PodResources) podresfilter.Result) string {
 	fp := podfingerprint.NewTracingFingerprint(len(podRes), st)
 	for _, pr := range podRes {
-		res := verifyFunc(pr)
-		if !res.Allow {
-			continue
+		if verifyFunc != nil {
```
ffromani: I'd prefer as null object pattern and remove these nil checks

ffromani (commented):
thanks for your PR. Unfortunately I'm a bit swamped these days so I'm beginning with a arch-level review. Few high level code shape comments:

pkg/numalocality/numalocality.go
shajmakh: is the problem in `pkg/podres` or in `pkg/podres/filter` because it's currently under `pkg/podres/numalocality`.
 would `pkg/numalocality` make more sense?

pkg/resourcemonitor/resourcemonitor.go
shajmakh: ack, will convert that to string for the policy only. in case any future additions needing other TM settings we can create a special struct then.

pkg/numalocality/numalocality.go
ffromani: it goes like this:
```
pkg - golang packages (outdated layout, but still)
`- podres - package(s) related to the podresources API
   `- filter - filter the items of a podresources API response based on the item properties
    `- numalocality - filter the items of podresources API based on the NUMA locality of the API response items
```
(on hindsight, lack of golang package docstrings is the real culprit, it's on me)

Therefore if we need some library code it needs to sit at `pkg` level like you mentioned, and that library code will be consumed by both the filter and your application code.

pkg/resourcemonitor/resourcemonitor.go
shajmakh: indeed, it was meant to be so by passing the AlwaysPass (no-op) function but I missed removing this check. added function description as well. Thanks for highlighting.

pkg/numalocality/numalocality.go
shajmakh: ack, thanks for elaborating. addressed in the latest push.

pkg/numalocality/numalocality.go
```diff
@@ -0,0 +1,93 @@
...
+			numaIDs.Insert(int(node.ID))
+		}
+	}
+	return numaIDs.UnsortedList()
+}
+
+// ResolveContainerSingleNUMANodeLocality Verifies that the container exclusive resources are all on a single NUMA node, if so it returns
+// the NUMA ID. If none of the exclusive resources are NUMA-pinned the function returns -1 and nil error. Multiple NUMA nodes are found for
+// the resources is a violation of the single-numa-node policy hence the function returns an error.
+func ResolveContainerSingleNUMANodeLocality(cnt *podresourcesapi.ContainerResources, coreIDToNodeIDMap map[int]int) (int, error) {
```
ffromani: I think we can simplify this significantly given the contract. Something like this (UNTESTED):
```
  func ResolveContainerSingleNUMANodeLocality(cnt *podresourcesapi.ContainerResources, coreIDToNodeIDMap map[int]int) (int, error) {
      if cnt == nil {
          return -1, fmt.Errorf("container is nil")
      }
      numaNodes := sets.New[int]()

      for _, cpuID := range cnt.CpuIds {
          nodeID, ok := coreIDToNodeIDMap[int(cpuID)]
          if !ok {
              return -1, fmt.Errorf("CPU ID %d not found in coreIDToNodeIDMap", cpuID)
          }
          numaNodes.Insert(nodeID)
      }
      for _, dev := range cnt.Devices {
          if len(dev.DeviceIds) == 0 {
              continue
          }
          numaNodes.Insert(GetNUMAIDs(dev.Topology)...)
      }
      for _, mem := range cnt.Memory {
          numaNodes.Insert(GetNUMAIDs(mem.Topology)...)
      }

      if numaNodes.Len() == 0 {
          return -1, nil
      }
      if numaNodes.Len() > 1 {
          return -1, fmt.Errorf("multiple NUMA nodes found for container resources: %v", numaNodes.UnsortedList())
      }
      return numaNodes.UnsortedList()[0], nil
  }
```

we can reintroduce more detailed error messages if needed by doing a check after each insertion (e.g. if after insertion len growth past 1, the resource just added is unaligned)

pkg/numalocality/numalocality.go
```diff
@@ -0,0 +1,93 @@
...
+		res, err = verifyResourceNUMAID("memory-"+mem.MemoryType, nodeIDs, res)
+		if err != nil {
+			return -1, fmt.Errorf("multiple NUMA nodes found for container resources: %w", err)
+		}
+	}
+
+	return res, nil
+}
+
+func verifyResourceNUMAID(resourceDesc string, resourceNUMAs []int, expectedNUMAID int) (int, error) {
```
ffromani: with my simplification this becomes unnecessary

pkg/resourcemonitor/resourcemonitor.go
```diff
@@ -261,21 +272,36 @@ func (rm *resourceMonitor) Scan(excludeList ResourceExclude) (ScanResponse, erro
 		return ScanResponse{}, err
 	}
 
-	respPodRes := resp.GetPodResources()
-	klog.V(6).Infof("resmon: podresources list: %s", collectPodsFromPodResources(respPodRes))
+	respRawPodRes := resp.GetPodResources()
```
ffromani, line 278: do we need this rename?

pkg/resourcemonitor/resourcemonitor.go
```diff
@@ -288,14 +314,32 @@ func (rm *resourceMonitor) Scan(excludeList ResourceExclude) (ScanResponse, erro
...
+		// NUMA placement attributes are worthy to be added only when PFP is enabled so that the
+		// when the consumer decode the data, it would rely on having a matching PFPs.
+		// numaplacement encoding is only done for pods that are eligible for NUMA placement (with
+		// exclusive resources), which are a subset of pods that are participating in PFP (it is
+		// not always the same pods as PFP because it depends on the filter function used)
+		metadata := unsupportedConfigurationForNumaPlacement
+		payload, err = ComputeNUMAPlacementPayload(numaEligiblePodRes, rm.tmPolicy, len(rm.topo.Nodes), rm.coreIDToNodeIDMap)
+		if err != nil {
+			klog.ErrorS(err, "resmon: while computing NUMA placement payload")
+			metadata = errorOccurredDuringNumaPlacementEncoding
```
ffromani: we need to encode the error differently. Should be a metadata sub-field in the `numaplacement` package or we need to omit the value. Replacing the expected value is not good enough

pkg/resourcemonitor/resourcemonitor.go
```diff
@@ -304,6 +348,17 @@ func (rm *resourceMonitor) Scan(excludeList ResourceExclude) (ScanResponse, erro
 			Resources: make(topologyv1alpha2.ResourceInfoList, 0),
 		}
 
+		// if there is only one NUMA node then adding the vector attribute is not needed because it will be the busiest
+		if payload != nil && payload.NUMANodes > 1 {
+			zoneVector, ok := payload.Vectors[nodeID]
```
ffromani, line 350: how do we skip setting the busiest node? this is not obvious to me

pkg/resourcemonitor/resourcemonitor.go
```diff
@@ -374,6 +429,69 @@ func (rm *resourceMonitor) Scan(excludeList ResourceExclude) (ScanResponse, erro
...
+			if numaNodeID == -1 {
+				// it's possible that a container does not belong to specific NUMA node (e.g. using shared pool resources)
+				//  in that case we skip it
+				continue
+			}
+
+			cntsToEncode = append(cntsToEncode, numaplacement.ContainerAffinity{
+				ID:       cntID,
+				NUMANode: numaNodeID,
+			})
```
ffromani: do we need an intermediate slice or can we just call `Encode` in the loop?

ffromani (commented):
thanks for your PR. Initial review.

pkg/resourcemonitor/resourcemonitor.go
shajmakh: for the busiest node there won't be anyhow any vector, and if the payload.NUMANodes is 1 means that there is only one NUMA, which is also true that it would be the busiest. we only add vector attributes for nodes that are not the besiest. We can rely on that numaplacement project to calculate the payload and can remove `&& payload.NUMANodes > 1` completely, we'll still have the same affect. would you prefer that?

ffromani: Makes sense, we're good here. Thanks!

pkg/resourcemonitor/resourcemonitor.go
shajmakh: I thought yes to differentiate between the raw unfiltered response and the filtered one(s)

pkg/numalocality/numalocality.go
shajmakh: Thanks for this, reading this I agree it simplifies the flow of this function, yet reduces the debuggability logs. I believe a container that messes up the numaplacement encoding deserves to be logged with as much reasonable details as possible, which is what `verifyResourceNUMAID` integration is serving. We can reject the container if it belongs to multi-numas, but since we don't have extra input in the logs for "why so?", we'll need to figure this out ourselves.  `verifyResourceNUMAID` would give us more details about that, especially with modern hardware where resources do not have specific numa alignement, and it is there as a function to avoid duplicating the same checks for all the container resources (cpus, memory, devices)

ffromani: Variant restoring the detailed error messages
```
  func ResolveContainerSingleNUMANodeLocality(cnt *podresourcesapi.ContainerResources, coreIDToNodeIDMap map[int]int) (int, error) {
        if cnt == nil {
                return -1, fmt.Errorf("container is nil")
        }

        numaNodes := sets.New[int]()

        for _, cpuID := range cnt.CpuIds {
                nodeID, ok := coreIDToNodeIDMap[int(cpuID)]
                if !ok {
                        return -1, fmt.Errorf("CPU ID %d not found in coreIDToNodeIDMap", cpuID)
                }
                numaNodes.Insert(nodeID)
                if numaNodes.Len() > 1 {
                        return -1, fmt.Errorf("cpu-%d is pinned to NUMA %d, conflicting with %v", cpuID, nodeID, numaNodes.UnsortedList())
                }
        }

        for _, dev := range cnt.Devices {
                if len(dev.DeviceIds) == 0 {
                        continue
                }
                nodeIDs := GetNUMAIDs(dev.Topology)
                numaNodes.Insert(nodeIDs...)
                if numaNodes.Len() > 1 {
                        return -1, fmt.Errorf("device %q is pinned to NUMAs %v, conflicting with %v", dev.ResourceName, nodeIDs, numaNodes.UnsortedList())
                }
        }

        for _, mem := range cnt.Memory {
                nodeIDs := GetNUMAIDs(mem.Topology)
                numaNodes.Insert(nodeIDs...)
                if numaNodes.Len() > 1 {
                        return -1, fmt.Errorf("memory %q is pinned to NUMAs %v, conflicting with %v", mem.MemoryType, nodeIDs, numaNodes.UnsortedList())
                }
        }

        if numaNodes.Len() == 0 {
                return -1, nil
        }
        return numaNodes.UnsortedList()[0], nil
  }
```

The two main reasons I'm suggesting this approach are:
1. your code was passing the context data (`res`) back and forth in helper _which also had the responsability to mutate it_, not just check it. Plus, `res` was carrying 2 signals: a. are the NUMA resources aligned and b. if so, where, which at the return value we decompose in (int, error) because we acknowledge these are actually 2 signals: a integer (placement) and a flag (bool or error). Now we carry the data structure once across the flow and we extract the signal once and at the end
2. the resource classes are similar but yet different enough which require different treatment to build a good error. Forcing them in a common shape (resource name as string) creates accidental complexity - we can just capture the differences when building an error, which we need to do anyway. The repeated code would just the `if numaNodes.Len()` boilerplate, too shallow for a helper I believe.

pkg/resourcemonitor/resourcemonitor.go
shajmakh: we can call encode per container indeed, though the slice is there to manage to print the participating containers in one shot so we would get this in the logs (instead of a line per container) similarly to PFP log:
```
I0608 11:51:24.999234       1 resourcemonitor.go:315] resmon: pfp: > processing node "cnfdr8.telco5g.eng.rdu2.redhat.com"
> processing 2 pods
+ numaresources/tas-4
+ numaresources/tas-1
= pfp0v00162519433151dc53e
I0608 11:51:24.999256       1 resourcemonitor.go:445] resmon: collecting containers that are eligible for NUMA placement
I0608 11:51:24.999308       1 resourcemonitor.go:483] "resmon: encoding containers NUMA affinity" containers=<
	numaresources/tas-4/app-exclusive-1 -> 0
	numaresources/tas-1/app-exclusive-1 -> 0
 > payload="npv0v001::ve=leb89::cc=2::nn=2::bn=0"

```

pkg/numalocality/numalocality.go
shajmakh: > is pinned to NUMAs %v, conflicting with %v", dev.ResourceName, nodeIDs, numaNodes.UnsortedList())

would look like:
"is pinned to NUMAs [1,0], conflicting with [0,1]"
`[1,0]` is same as [0,1] then why conflicting? and what is the expected? one wouldn't know to answer these questions without looking at the code.

regarding point (1), verifyResourceNUMAID is not mutating `res`, we decide to mutate it in `ResolveContainerSingleNUMANodeLocality` but it would anyhow not make sense to keep going the function flow if it returned an error. 
as for (2), the new variant still kept the same error logs so why not to use the common shape?

pkg/resourcemonitor/resourcemonitor.go
ffromani: the right fix IMO here is pushing the log layer into the `numaplacement` package, like we did for `NewTracingFingerprint` in the `podfingerprint` package

pkg/numalocality/numalocality.go
shajmakh: I think only the name `verifyResourceNUMAID` is confusing without a description. would adding this make it a bit more clear?
```
verifyResourceNUMAID checks if the resource with the specified name belongs to single NUMA ID and is equal to expectedNUMAID
```

pkg/resourcemonitor/resourcemonitor.go
shajmakh: yes indeed, that sounds better, I'll address that. can this modification be handled in a different follow-up PR?

pkg/numalocality/numalocality.go
ffromani: Is this function meant to be called before or after we ensured the kubelet is running with the topology manager `single-numa-node`? because if we can depend on that, then by construction we will only get both a single NUMA value _AND_ that value would be guaranteed consistent by the kubelet, so we don't actually need all this complexity: just fetch the resource class easier to access, fetch the NUMA locality and move on.

> regarding point (1), verifyResourceNUMAID is not mutating res, we decide to mutate it in ResolveContainerSingleNUMANodeLocality but it would anyhow not make sense to keep going the function flow if it returned an error.

but the function didn't just check the value: it did check _and possibly return_ a mutated value the caller has
to carry around. Which works, but pushes the state back and forth. In some cases this is unavoidable (easiest example: builtin `append()`), in this case I'm not sold it is unavoidable.

>as for (2), the new variant still kept the same error logs so why not to use the common shape?

because the common shape emerges naturally with the error value; forcing prematurely seems an artifact, which we're exploring if we can avoid.

pkg/resourcemonitor/resourcemonitor.go
ffromani: yes. Drop the detailed logs for now, we can start merging the base feature.

pkg/numalocality/numalocality.go
shajmakh: this is meant to run after ensuring kubelet uses `single-numa-node`, it was extracted in commit ebc48da610d16c7ca5e40e75c98b33757c23db83 just to have control over the unit tests but looks like it backfired with unnecessary complexity. I'm dropping that commit, since the function is only used in one place which is guaranteed to use a single NUMA only.

pkg/resourcemonitor/resourcemonitor.go
shajmakh: metadata is built from the payload, numaplacement package assumes single-numa affinity (https://github.com/k8stopologyawareschedwg/numaplacement#numaplacement-efficient-per-numa-container-placement-encoding) and since we are the consumers of the encoding, and we feed the encoder with the containers, what we can so is add a err field to payload, and upon packing the metadata from payload, if `payload.EncodingError !=""` it will also append that subfield in the metadata string. Is that a reasonable way to encode the errors?

pkg/resourcemonitor/resourcemonitor_test.go
```diff
@@ -1360,7 +1360,27 @@ func TestResourcesScan(t *testing.T) {
...
+	rm, err := NewResourceMonitor(
+		Handle{PodResCli: mockPodResClient},
+		Args{},
+		tmPolicy,
+		WithNodeName("TEST"),
+		WithTopology(&topo),
+		WithK8sClient(fake.NewSimpleClientset()),
+	)
+	assert.NoError(t, err)
+	assert.Equal(t, tmPolicy, rm.tmPolicy)
```
ffromani: we should never test private fields.
A good way forward could be to expose a method like `Enforcing() bool` which returns true iff the tmPolicy is `single-numa-node`. The reason is that in the past and current kubelet, resources have a NUMA affinity only if thy are devices (self-evident, always true) OR if the policy is single-numa-node

pkg/podres/filter/numalocality/numalocality.go
```diff
@@ -81,15 +82,5 @@ func Required(pr *podresourcesapi.PodResources) bool {
...
-	}
-	// if Nodes is not given, this means "don't care about locality". It's a legal representation.
-	for _, node := range topo.Nodes {
-		// setting node.ID == -1 is also a legal representation for "don't care about locality".
-		if node.ID >= 0 {
-			return true
-		}
-	}
-	return false
+	return len(numalocality.GetNUMAIDs(topo)) > 0
```
ffromani: just inline and remove it. `IsPresent` was the old API. new code should use `Verify`. Let's take this chance to remove some cruft.

pkg/resourcemonitor/resourcemonitor.go
```diff
@@ -290,14 +298,153 @@ func (rm *resourceMonitor) Scan(excludeList ResourceExclude) (ScanResponse, erro
...
+								}
+
+								if len(nodeIDs) > 1 {
+									// should never happen on singleNUMANode policy
+									return -1, fmt.Errorf("multiple NUMA nodes found for container %s", cnt.Name)
+								}
+								return nodeIDs[0], nil
+							}
+							return -1, nil
+						}
```
ffromani: this is not really inline so let's move this helper as private function in this file

pkg/resourcemonitor/resourcemonitor.go
shajmakh: https://github.com/k8stopologyawareschedwg/numaplacement/issues/5

pkg/resourcemonitor/resourcemonitor.go
shajmakh: https://github.com/k8stopologyawareschedwg/numaplacement/issues/4

pkg/resourcemonitor/resourcemonitor.go
ffromani: this helper can probably survive the extraction you do in a followup (need to check)

pkg/podres/filter/numalocality/numalocality.go
shajmakh: > IsPresent was the old API. new code should use Verify.

each is for different goal, Verify uses IsPresent under the hood, new code would use `len(numalocality.GetNUMAIDs(topo)) > 0` instead. I see Required is deprecated, I'll remove both.

pkg/resourcemonitor/resourcemonitor.go
ffromani: If we need to encode the error at all, then yes, this seems like a direction worth exploring. But I'm thinking harder if encoding the error brings any value vs omitting the data at all on error path.

pkg/resourcemonitor/resourcemonitor_test.go
shajmakh: why to limit that to single-numa-node check, I'm in favor of making it more flexible to verify if the field is a specific value. would that make sense?

ffromani: because that value is the only TM policy which enables numa locality for CPU and memory. Devices have always a NUMA affinity, but without `single-numa-node` we can't do precise accounting of resources. This limitation / assumption is already everywhere in the stack.

pkg/resourcemonitor/resourcemonitor.go
```diff
@@ -252,6 +254,10 @@ func WithNodeName(name string) func(*resourceMonitor) {
 	}
 }
 
+func (rm *resourceMonitor) HasTopologyManagerPolicy(policy string) bool {
+	return rm.tmPolicy == policy
+}
```
ffromani, line 266: this seems unnecessary generalization, but is not a big deal now. We can move forward with this.

pkg/resourcemonitor/resourcemonitor.go
```diff
@@ -40,18 +40,25 @@ import (
 	ghwtopology "github.com/jaypipes/ghw/pkg/topology"
 	"github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/topology/v1alpha2"
 	topologyv1alpha2 "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/topology/v1alpha2"
+	"github.com/k8stopologyawareschedwg/numaplacement"
 	"github.com/k8stopologyawareschedwg/podfingerprint"
 
 	"github.com/k8stopologyawareschedwg/resource-topology-exporter/pkg/metrics"
+	"github.com/k8stopologyawareschedwg/resource-topology-exporter/pkg/numalocality"
```
ffromani: nit: I'd have renamed this import like `numaloclib`, but it's fine. Please don't resubmit for this.

pkg/resourcemonitor/resourcemonitor.go
```diff
@@ -40,18 +40,25 @@ import (
...
 )
 
 const (
 	defaultPodResourcesTimeout = 10 * time.Second
 	// obtained these values from node e2e tests : https://github.com/kubernetes/kubernetes/blob/82baa26905c94398a0d19e1b1ecf54eb8acb6029/test/e2e_node/util.go#L70
+
+	TopologyManagerPolicySingleNUMANode = "single-numa-node"
+
+	unsupportedConfigurationForNumaPlacement = "unsupported"
+	errorOccurredDuringNumaPlacementEncoding = "error occurred"
```
ffromani, line 61: we need to remove this error handling. If the numaplacement encoder fails to construct numaplacement data, we should _not_ set values in the NRT objects. The consumer must tolerate missing or disappearing numaplacement data.

We don't have means to provide better semantics anyway (https://github.com/k8stopologyawareschedwg/numaplacement/issues/4#issuecomment-4649743287)

pkg/resourcemonitor/resourcemonitor.go
```diff
@@ -294,14 +302,153 @@ func (rm *resourceMonitor) Scan(excludeList ResourceExclude) (ScanResponse, erro
 		klog.V(6).Infof("resmon: pfp: %s", st.Repr())
 
 		podfingerprint.MarkCompleted(st)
+
+		// computing payload is valuable only on singleNumaNode policy because only under that
+		// condition there will be 1:1 stable mapping between the containers to the NUMA nodes.
+		// NUMA placement attribute is worthy to be added only when PFP is enabled so that the
+		// when the consumer decode the data, it would rely on having a matching PFPs.
+		metadata := unsupportedConfigurationForNumaPlacement
```
ffromani: this is one of the many examples why we don't need this generalization: https://github.com/k8stopologyawareschedwg/resource-topology-exporter/pull/396/changes/ff93c3dc7f50e14a014273066599fc1fc24ab739#r3375511826

pkg/resourcemonitor/resourcemonitor.go
```diff
@@ -294,14 +302,153 @@ func (rm *resourceMonitor) Scan(excludeList ResourceExclude) (ScanResponse, erro
...
+		if rm.tmPolicy == TopologyManagerPolicySingleNUMANode {
+			// we are only interested in pods that are eligible for NUMA placement, which has exclusire resources requests;
```
ffromani: typo: exclusive

pkg/resourcemonitor/resourcemonitor.go
```diff
@@ -274,22 +275,36 @@ func (rm *resourceMonitor) Scan(excludeList ResourceExclude) (ScanResponse, erro
...
+	for _, pr := range respRawPodRes {
+		nl := numalocfilter.Verify(pr)
+		if !nl.Allow {
+			continue
+		}
+		sb.WriteString(sep + pr.Namespace + "/" + pr.Name + "/" + nl.Ident + "=" + nl.Reason)
+		sep = "  "
+		numaEligiblePodRes = append(numaEligiblePodRes, pr)
+	}
+	klog.V(6).Infof("resmon: numa eligible podresources list: %s", sb.String())
```
ffromani, line 293: this pattern always builds a long string and optionally logs it, in one go which makes the parsing harder. it's better to use contextual logging here: log inside the loop using a flow key to make it easy to filter all the log lines pertaining to the flow. Can be fixed later.

pkg/resourcemonitor/resourcemonitor.go
```diff
@@ -573,22 +579,19 @@ func computePodFingerprintFromPodResources(podRes []*podresourcesapi.PodResource
...
-		sb.WriteString("  " + pr.Namespace + "/" + pr.Name + desc)
+		sb.WriteString(sep + pr.Namespace + "/" + pr.Name)
+		sep = "  "
 	}
 	val := sb.String()
 	if len(val) == 0 {
 		return ""
 	}
-	return val[2:]
+	return val
```
ffromani, line 590: I'm not sure this way is simpler or cheaper than the previous, but not a big deal now

ffromani (commented):
basically we need to stop adding the numaplacement attribute on error (https://github.com/k8stopologyawareschedwg/numaplacement/issues/4#issuecomment-4649743287). For observability, besides logs (at V=2 level) we can think about adding/changing conditions or else, and can be done later.

pkg/resourcemonitor/resourcemonitor.go
shajmakh: the only value I see in keeping numaplacement error reporting is a quick way to point out possible initial reasons for consumer functionality failures without the need to look into the logs or track the build if it supports the numaplecament encoding or not. do we think this is worth it? do we have any u/s objection or something against the  to reporting the error in the API?

ffromani: debugging is a very legitimate concern and worth adding, but the fix is misplaced. I think we should add a new condition. `pkg/podreadiness`. Encoding numaplacement is not (yet) critical to the exporter operation, so we should NOT set a readiness gate on it. Can be done in a followup PR

shajmakh: yes, I think this requires more discussion so I'll let this be done in a different PR. `pkg/podreadiness` seems too broad for numaplacement though.

pkg/resourcemonitor/resourcemonitor_test.go
shajmakh: okay I see, I can handle that while still gate the numaplacement encoding when the consition is false. will address it.

pkg/resourcemonitor/resourcemonitor.go
ffromani: I think the change won't be too big - just surfacing the condition. It can totally be done on a separate PR.

pkg/resourcemonitor/resourcemonitor.go
shajmakh: it wasn't clear at the first moment why we are quitting the first 2 chars, this was meant to clarify this point only without a change in the output

ffromani: we can keep this variant, no big deal, exactly because I can't say it's better or worse than before

ffromani: The previous version avoided per-iteration branching, which is a nice property in general, but in this context doesn't make much of a difference. Everything else equal I'd prefer to keep the old one but please don't resubmit just for this.

pkg/resourcemonitor/resourcemonitor.go
shajmakh: ack, removed for now. Thanks.

pkg/resourcemonitor/resourcemonitor.go
shajmakh: understood, removed for now so that the metadata would only be visible if it is valid

pkg/resourcemonitor/resourcemonitor.go
shajmakh: followed the same pattern used to print out the raw podres list. will improve this in the next PR.

pkg/resourcemonitor/resourcemonitor.go
shajmakh: done

pkg/resourcemonitor/resourcemonitor.go
```diff
@@ -40,18 +40,25 @@ import (
...
 	"github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/topology/v1alpha2"
 	topologyv1alpha2 "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/topology/v1alpha2"
+	"github.com/k8stopologyawareschedwg/numaplacement"
 	"github.com/k8stopologyawareschedwg/podfingerprint"
 
 	"github.com/k8stopologyawareschedwg/resource-topology-exporter/pkg/metrics"
+	numaloclib "github.com/k8stopologyawareschedwg/resource-topology-exporter/pkg/numalocality"
 	podresfilter "github.com/k8stopologyawareschedwg/resource-topology-exporter/pkg/podres/filter"
-	"github.com/k8stopologyawareschedwg/resource-topology-exporter/pkg/podres/filter/numalocality"
+	numalocfilter "github.com/k8stopologyawareschedwg/resource-topology-exporter/pkg/podres/filter/numalocality"
```
ffromani, line 49: this rename is now useless, but no bother

pkg/resourcemonitor/resourcemonitor.go
ffromani: I still see constants, these should be removed as well

pkg/resourcemonitor/resourcemonitor.go
```diff
@@ -304,6 +345,17 @@ func (rm *resourceMonitor) Scan(excludeList ResourceExclude) (ScanResponse, erro
 			Resources: make(topologyv1alpha2.ResourceInfoList, 0),
 		}
 
+		// if there is only one NUMA node then adding the vector attribute is not needed because it will be the busiest
+		if payload != nil && payload.NUMANodes > 1 {
+			zoneVector, ok := payload.Vectors[nodeID]
```
ffromani, line 350: we can use a method in `Payload` in the `numaplacement` package:
```
func (pl *Payload) GetVector(nodeID int) (string, bool) {
    if pl == nil {
        return "", false
    }
    return payload.Vectors[nodeID]
}
```

ffromani (approved):
LGTM

please address the pending/open points with followup PRs.

pkg/resourcemonitor/resourcemonitor.go
shajmakh: addressed in https://github.com/k8stopologyawareschedwg/resource-topology-exporter/pull/397/changes/2a40531a90fd259119bf7c26b162d7b5fd9c55df

PR #397, opened 2026-06-09
ctxlog: port ResourceMonitor to contextual logging

ffromani:
begin the transition to contextual logging by porting the `resourcemonitor` package.
This enables us to remove some technical debt (lack of `context` wiring) and simplify the logging.
We will continue with the contextual logging transition with more followup PRs.

### DIFF
--- a/pkg/resourcetopologyexporter/resourceobserver.go
+++ b/pkg/resourcetopologyexporter/resourceobserver.go
@@ -1,6 +1,7 @@
 package resourcetopologyexporter
 
 import (
+	"context"
 	"fmt"
 	"time"
 
@@ -26,7 +27,7 @@ type ResourceObserver struct {
 
 func NewResourceObserver(hnd resourcemonitor.Handle, args resourcemonitor.Args, tmPolicy string) (*ResourceObserver, error) {
 	resMon := resourcemonitor.NewResourceMonitor(hnd, args, tmPolicy)
-	if err := resMon.Setup(); err != nil {
+	if err := resMon.Setup(context.TODO()); err != nil {
 		return nil, fmt.Errorf("failed to setup ResourceMonitor: %w", err)
 	}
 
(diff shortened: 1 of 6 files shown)


pkg/resourcetopologyexporter/resourceobserver.go
```diff
@@ -26,7 +27,7 @@ type ResourceObserver struct {
 
 func NewResourceObserver(hnd resourcemonitor.Handle, args resourcemonitor.Args, tmPolicy string) (*ResourceObserver, error) {
 	resMon := resourcemonitor.NewResourceMonitor(hnd, args, tmPolicy)
-	if err := resMon.Setup(); err != nil {
+	if err := resMon.Setup(context.TODO()); err != nil {
```
shajmakh: why are we not passing the top level context here?

pkg/resourcetopologyexporter/resourceobserver.go
```diff
@@ -59,7 +60,7 @@ func (rm *ResourceObserver) Run(eventsChan <-chan notification.Event, condChan c
 			metrics.UpdateWakeupDelayMetric(monInfo.UpdateReason(), float64(tsWakeupDiff.Milliseconds()))
 
 			tsBegin := time.Now()
-			scanRes, err := rm.resMon.Scan(rm.resourceExclude)
+			scanRes, err := rm.resMon.Scan(context.TODO(), rm.resourceExclude)
```
shajmakh: same here, why not construct the context in main and pass it through the sub-processes?

pkg/resourcetopologyexporter/resourceobserver.go
ffromani: because it will cause cascading changes. But let me check how'd it turn out.

pkg/resourcetopologyexporter/resourceobserver.go
ffromani: same (you probably meant call tree?)

shajmakh (approved):
LGTM

PR #398, opened 2026-06-11
ctxlog: wire and use context the in the resourcetopologyexporter package

ffromani:
We can and should get the logger from the context. Includes some minor resourcemonitor fixes.


shajmakh (approved):
LGTM

PR #399, opened 2026-06-11
cleanup updates following #396

shajmakh:
See commits for details.


pkg/resourcemonitor/resourcemonitor.go
```diff
@@ -441,19 +441,19 @@ func ComputeNUMAPlacementPayload(logger logr.Logger, numaEligiblePodRes []*podre
 	// computing payload is valuable only on singleNumaNode policy because only under that
 	// condition there will be 1:1 stable mapping between the containers to the NUMA nodes.
 	if topologyManagerPolicy != TopologyManagerPolicySingleNUMANode {
-		logger.V(4).Info("topology manager policy not supported for NUMA placement", "policy", topologyManagerPolicy)
+		logger.V(2).Info("topology manager policy not supported for NUMA placement", "policy", topologyManagerPolicy)
```
ffromani: thinking more about it, please drop this commit. I think the logs makes more sense as-is rather than as-suggested.

ffromani (commented):
please drop a commit as suggested inline and fix the linter warnings and we can merge

PR #400, opened 2026-06-30
resmon: detect and expose non-topology-aware resources via NRT attribute

Tal-or:
Some extended resources on a node are not tied to any NUMA node -- either because the device plugin reports them without topology information, or because they aren't backed by a device plugin at all.

Today RTE silently ignores these resources, making them invisible to topology-aware schedulers.

This PR detects both kinds and exposes their names as a new top-level NRT attribute `nodeNonTopologyResources`, so scheduler-plugins can identify and exclude them from NUMA-aware scheduling decisions.

### DIFF
--- a/pkg/resourcemonitor/resourcemonitor.go
+++ b/pkg/resourcemonitor/resourcemonitor.go
@@ -443,6 +443,22 @@ func (rm *resourceMonitor) Scan(ctx context.Context, excludeList ResourceExclude
 
 		zones = append(zones, zone)
 	}
+
+	if rm.nonTopologyResources.Len() > 0 {
+		resNames := make([]string, 0, rm.nonTopologyResources.Len())
+		for resName := range rm.nonTopologyResources {
+			if inExcludeSet(excludeSet, resName, rm.nodeName) {
+				continue
+			}
+			resNames = append(resNames, resName.String())
+		}
+		sort.Strings(resNames)
+		scanRes.Attributes = append(scanRes.Attributes, topologyv1alpha2.AttributeInfo{
+			Name:  AttributeNodeNonTopologyResources,
+			Value: strings.Join(resNames, ","),
+		})
+	}
+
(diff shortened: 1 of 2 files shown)


pkg/resourcemonitor/resourcemonitor.go
```diff
@@ -575,9 +598,51 @@ func (rm *resourceMonitor) updateNodeResources(ctx context.Context) error {
...
+		if err != nil {
+			logger.V(1).Info("error fetching local node, continuing without non-device-plugin resource detection", "err", err)
+			return nil
+		}
+	}
+	rm.detectNonDevicePluginResources(logger, node)
 	return nil
 }
 
+func (rm *resourceMonitor) getLocalNode(ctx context.Context) (*v1.Node, error) {
```
ffromani, line 613: this solution must not require us listing nodes, even once: that would increase the RBAC surface.
If this is the only solution, we can talk about it. But in our case, we can just report the extended resources backed by device plugins - things to include, not to exclude.

Tal-or: From what I saw on NROP we have List and Watch rbac on the node for RTE (but not get)

I may overlooked. Ill check again

Tal-or: I doubled checked and NROP provides `List` to RTE anyway. so no intended changes on RBAC.

ffromani: thanks. Let's avoid API calls if we can help it (and I think we can)

Tal-or: If we want to show the resources to include instead the ones we want to exclude, we can simply rely on the existing resources shown in the NRT - in NRT we only have resources backed by devices plugins + native resources.
IOW, it's possible that no changes are required on RTE side at all. let me check

